PR #3, opened 2021-01-22
API project setup

seerocode:
## Description

- Added base flask and modules setup
- Added virtual env and requirements file

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,11 +1,14 @@
 click==7.1.2
 Flask==1.1.2
 Flask-Cors==3.0.10
+Flask-SQLAlchemy==2.4.4
 gunicorn==20.0.4
+install==1.3.4
 itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
 nginx==0.0.1
 python-dotenv==0.15.0
 six==1.15.0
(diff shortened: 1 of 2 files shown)


api/requirements.txt
```diff
@@ -0,0 +1,14 @@
+click==7.1.2
+Flask==1.1.2
+Flask-Cors==3.0.10
+Flask-SQLAlchemy==2.4.4
+gunicorn==20.0.4
+install==1.3.4
+itsdangerous==1.1.0
+Jinja2==2.11.2
+MarkupSafe==1.1.1
+nginx==0.0.1
```
jnu: huh! didn't know there was an nginx python package. we're actually going to use nginx as a standalone binary inside the dockerfile. (there'll be a line in the docker file like `apt-get install nginx`). you can remove the nginx dependency here.

thanks for adding gunicorn! we'll set that up at the same time we set up nginx in the dockerfile

jnu (approved):
first pull request looks great 🎉  woo!!

just remove the nginx dep before merging

api/requirements.txt
seerocode: Removed Nginx!

seerocode:
Rebased main branch onto PR branch to pull in previous changes

PR #5, opened 2021-02-03
Database schema

jnu:
Add relational db schema.

The current version has a few issues based on feedback from BBC team:

 - Tags are not sufficiently generalized. We'd want Targets tracked for each tag. Tags can probably even take the place of things like Programs
 - Users can be part of multiple teams (are teams also just tags?)

Discussion about whether to scrap the relational DB and go with Mongo is still open -- if we don't feel good about hammering the above issues into the relational model, we should shift course.

If we keep this schema, we should make sure 1) indexes and nullability are set correctly for all the columns, and 2) we integrate with the flask-sqlalchemy library


api/database.py
```diff
@@ -0,0 +1,191 @@
...
+    category = Column(String(255), nullable=False)
+    category_value = Column(String(255), nullable=False)
+    
+    created = Column(DateTime, nullable=False)
+    updated = Column(DateTime)
+    deleted = Column(DateTime)
+
+
+if __name__ == '__main__':
+    engine = create_engine('mysql+mysqldb://root@localhost/bbc', echo=True)
```
jnu: could u change this to use postgres? the connection string should be very similar, but since we're also use FastAPI we can install an asyncio driver as well. more details about how to do that here: https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.asyncpg

api/requirements.txt
```diff
@@ -7,7 +7,8 @@ install==1.3.4
 itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
+mysqlclient==2.0.3
```
jnu: flagging this to remove and install postgres async driver per above

api/database.py
```diff
@@ -0,0 +1,191 @@
...
+    datasets = relationship('Dataset', secondary=dataset_tags,
+            back_populates='tags')
+    
+
+    created = Column(DateTime, nullable=False)
+    updated = Column(DateTime)
+    deleted = Column(DateTime)
+
+#polymorphic relationship 
+# class Tag_Associations(Base):
```
jnu: i'm not totally opposed to the polymorphic relationship idea, but i'm curious how easy it will be to fit into sqlalchemy's framework? have u found anything on the internet about this? i found this (old) article:
https://techspot.zzzeek.org/2007/05/29/polymorphic-associations-with-sqlalchemy/

api/database.py
```diff
@@ -0,0 +1,191 @@
...
+class Record(Base):
+    __tablename__ = 'record'
+
+    id = Column(Integer, primary_key=True)
+    dataset_id = Column(Integer, ForeignKey('dataset.id'), nullable=False)
+    publication_date = Column(DateTime)
+    
+    # category = Column(String(255), nullable=False)
+    # category_value = Column(String(255), nullable=False)
+    # TODO - this is if we're entering aggregates. Individual level data we
```
jnu: Yea we will certainly have to update this as we determine how we will end up storing data!

api/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+python3.8 -m venv venv
+```
+
+## 2) Set up database (TODO!!)
+
+Do this the first time you set up the project.
+
+To play with the schema I'm just using mysql installed via homebrew on my mac. User is root with no password. You can manually run `create database bbc;`, then run `python database.py` after step (4) below to create tables. [CHANGE ME LATER]
```
jnu: could u change this to say postgres?

api/database.py
```diff
@@ -0,0 +1,191 @@
...
+    updated = Column(DateTime)
+    deleted = Column(DateTime)
+
+#polymorphic relationship 
+# class Tag_Associations(Base):
+#     __tablename__ = 'tag_associations'
+#     id = Column(Integer, primary_key=True)
+#     tag_id = Column(Integer)
+#     table_type = Column(String(255), nullable=False)
+#     record_id = Column(Integer)
```
jnu: we should always add indexes to columns that are gonna be used for joins. the reason is, the database has to match records from one table to another to join them, and that process is very slow if it hasn't already indexed all the columns it is comparing.

i'd guess the polymorphic approach would be a little less performant than using separated mapping tables since you can't use a foreign key. as long as the join column is indexed it should be ok tho

jnu: (lmk if you have any more questions about that!)

jnu (commented):
this looks like a great start!! 

i made a couple small suggestions to change. I think the polymorphic tag idea might be more complicated than just using the different mapping tables. my thinking is that SqlAlchemy makes managing extra columns and tables reasonably easy, so it's not as critical to condense things into a single mapping table. what do u think?

api/requirements.txt
nmanduj1: will do!

api/database.py
nmanduj1: ill take a look- might not be worth incorporating if it makes our lives harder

nmanduj1:
hey joe, sorry you had to go through and check everything- i was just hoping for feedback on the schema. thanks for reviewing everything, tho! I will update this pr tomorrow! hope you have a good night!

api/database.py
```diff
@@ -0,0 +1,163 @@
...
+    deleted = Column(DateTime)
+
+
+class Record(Base):
+    __tablename__ = 'record'
+
+    id = Column(Integer, primary_key=True)
+    dataset_id = Column(Integer, ForeignKey('dataset.id'), nullable=False, index=True)
+    publication_date = Column(DateTime)
+    category = Column(String(255), nullable=False)
```
jnu, line 152: Just noting (no need to change anything) for `category` we might store some crosstabs and store them, like `gender,race` for category and `female,black` for category_value, depending on what we're allowed to compute.

jnu (commented):
looks great to me! i can't `approve` bc technically i created this PR, but it should be set to merge!!

nmanduj1 (approved):

PR #7, opened 2021-02-25
Implement graphlq ariadne

nmanduj1:


api/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+app = Flask(__name__)
+
+# leaving in as placeholder
+# app.config["SQLALCHEMY_DATABASE_URI"] = ""
+# db = SQLAlchemy(app)
+
+@app.route('/')
+def hello():
+    return 'Hello!'
```
jnu: 👋

api/app.py
```diff
@@ -1,12 +1,36 @@
...
-def say_hello_world():
-    return "Hello, world!"
+query = ObjectType("Query")
+mutation = ObjectType("Mutation")
 
-if __name__ == '__main__':
-    app.run(debug=True)
\ No newline at end of file
+query.set_field("users", resolve_users)
+mutation.set_field("createUser", resolve_create_user) # feels like a camelCase vs under_score headache is looooming-
```
jnu: ah I see -- i feel totally fine with having camelCase in GraphQL and snake_case in python in this way (since the camelCase is just in a string)

but it looks like ariadne also makes use of decorators for more convenient syntax. i'm going to read the docs and see if i can understand what they're trying to show

api/queries.py
```diff
@@ -0,0 +1,13 @@
+from ariadne import convert_kwargs_to_snake_case
+
+@convert_kwargs_to_snake_case
+def resolve_users(obj, info):
```
jnu, line 8: based on the Ariadne docs, it looks like maybe the organization pattern here should be to put the `ObjectType("Query")` in this file, so that you don't have to import each resolver into app.py (it'll be a lot!)

this will also let you use the decorator syntax that'll be a little tidier:

in queries.py:
```py
query = ObjectType("Query")

@query.field("users")
def resolve_users(obj, info):
    ...
```

in app.py:
```py
from queries import query
...
make_executable_schema(type_defs, query)
```

(and same for mutations)

jnu (approved):
looks great to me!! nice work!!!

api/app.py
```diff
@@ -17,20 +20,11 @@
...
 
-    success, result = graphql_sync(
-        schema,
-        data,
-        context_value=request,
-        debug=app.debug
-    )
+@app.get("/")
+def home():
+    return "Ahh!! Aliens!"
```
seerocode, line 22: I vote this is the new "Hello, World!"

api/app.py
```diff
@@ -1,12 +1,30 @@
...
-if __name__ == '__main__':
-    app.run(debug=True)
\ No newline at end of file
+app = FastAPI()
+
+query = ObjectType("Query")
+mutation = ObjectType("Mutation")
+
+query.set_field("users", resolve_users)
+mutation.set_field("createUser", resolve_create_user) # feels like a camelCase vs under_score headache is looooming-
```
seerocode: Does ariadne require the string to be camel case?

jnu (approved):
looks great nice work!!!! just add a smol section in `api/README.md` with instructions about how to run the app for developing!

jnu:
btw i tested it out with running `uvicorn app:app --reload` and it looks great!

PR #8, opened 2021-03-01
BBC UI components setup

seerocode:
Components added to set up UI for future content and pages:
- Sidebar with submenu for datasets and stats
- Header nav for avatar and user's name (with bottom border image)
- Middle content pane
- Footer


seerocode:
Alright, friends!! We are at this point:

desktop -

client/public/manifest.json
```diff
@@ -1,21 +1,16 @@
 {
-  "short_name": "React App",
-  "name": "Create React App Sample",
+  "short_name": "50:50 Reporting Tool",
+  "name": "50:50 The Equality Project Reporting Tool",
```
jnu: this is good for now but we should brainstorm more general project names internally so it's not tied directly to bbc!

client/public/manifest.json
```diff
@@ -1,21 +1,16 @@
...
   "icons": [
     {
       "src": "favicon.ico",
       "sizes": "64x64 32x32 24x24 16x16",
       "type": "image/x-icon"
     },
     {
-      "src": "logo192.png",
+      "src": "5050logo.png",
```
jnu: we'll probably want to generalize this to be "Brandable" with different deployments. No need to do now, but i'll make a story in clubhouse to follow up on this.

jnu: https://app.clubhouse.io/stanford-computational-policy-lab/story/81/make-reporting-tool-brandable

client/src/components/AppHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+                style={{ float: "right", lineHeight: "64px" }}
+              >
+                <Menu.Item
+                  disabled
+                  key="userProfileImage"
+                  className="header__user-profile-item"
+                >
+                  <Avatar
+                    size="large"
+                    style={{ backgroundColor: "#87d068" }}
```
jnu, line 27: it'll be easier to theme the app later if right off the bat we use variables everywhere for style items.

there are different theming libraries for react -- we can use one of those, or the simplest thing is just to create a module with some constants in it

client/src/components/AppHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default class AppHeader extends React.Component {
+  render() {
+    return (
+      <div>
+        <Header className="header">
+          <Row wrap={false}>
+            <Col flex="none">
+              <div className="header__logo">
+                <img src={logo} alt="App logo" />
```
jnu, line 17: thanks for adding alt text!! this is great for accessibility. We will also need to add the other accessibility attributes, like the [aria- tags](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA).

We will _also_ need to add internationalization (remember the program/programme variation). like accessibility, this will be a pain to add later, but not so bad if we build it in from the start. see https://react.i18next.com/ for a reasonable library

client/src/App.tsx
```diff
@@ -1,25 +1,30 @@
...
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+    <Layout style={{ height: "100vh" }}>
+      <AppHeader />
+      <Layout>
+        <AppSidebar />
+        <Layout style={{ padding: "0 24px 24px" }}>
```
jnu, line 15: just curious, did you copy these margins vs paddings from some quick-start guide? it's fine as is, but mixing padding/margins like this can be finicky (especially when u use top margins, which can get collapsed, and sometimes that's what you want but sometimes not!)

jnu (commented):
great start! soon (not necessarily in this pr) we should add accessibility testing tools and also internationalization tools

client/public/manifest.json
seerocode: Agreed!

client/public/manifest.json
seerocode: I was wondering if we could do something with Yaml or some JSON object to add to whatever theme we use that's specific to the org.

client/src/App.tsx
seerocode: Yeah this is from the ant design library. Seemed arbitrary 🤷‍♀️

client/src/components/AppHeader.tsx
seerocode: What do you mean by variables? Similar to the comment above?

client/src/components/AppHeader.tsx
seerocode: I found this too I'm going to take a look at now: https://github.com/dequelabs/axe-core-npm/tree/develop/packages/react

Plus: 
- https://github.com/dequelabs/axe-core (same as above)
- https://github.com/nickcolley/jest-axe

Also seems like ant design library hasn't addressed accessibility issues yet. I did a Wave check and a couple of issues came up. Here's the antd issue for accessibility: https://github.com/ant-design/ant-design/issues/22343

client/src/components/AppHeader.tsx
seerocode: Not sure if this relevant (for themes, at least): https://css-tricks.com/theming-and-theme-switching-with-react-and-styled-components/

Maybe an extension of this: https://www.pluralsight.com/guides/how-to-store-and-read-configuration-files-using-react

client/public/manifest.json
seerocode: @jnu  how about a generic "Media Equality Reporting Tool"? not sure if app is more appropriate

seerocode:
Fixed the avatar! 
image

client/public/manifest.json
jnu: sounds good to me!

client/public/manifest.json
jnu: that might be part of it, but it's annoying to do images that way.

it seems like we should use docker to handle a lot of the theming. to me it makes most sense to incorporate this into the docker build process, so that we can do something like:

```
docker build -t bbc5050:latest . --build-arg LOGO=5050logo.png --build-arg STYLES=5050stylesheet.css
```

what do you think of that?

seerocode:
@jnu is this OK to merge for now?

client/src/components/AppHeader.tsx
jnu: yea the theming and theme switching one seems relevant, but maybe overkill

looks like this is what antd suggests for theming https://ant.design/docs/react/customize-theme

jnu (approved):
looks good!! great start, thank u!!

client/public/manifest.json
seerocode: Didn't know we could do that! That sounds good. What about if there are multiple assets besides the logo? @jnu

client/src/components/AppHeader.tsx
seerocode: Thanks! I like their less file option so I'll add that to the story for theming: https://ant.design/docs/react/customize-theme#Customize-in-less-file

PR #11, opened 2021-03-09
[CH82] Support Internationalization

seerocode:
### Description

- Added library [React i18n](https://github.com/i18next/react-i18next) library for internationalization support that includes browser language detector and lazy loading translations 
- Added locale for en-GB to support UK English and custom translations


seerocode:
Missing tests

client/src/App.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Suspense } from "react";
```
jnu, line 1: 😲 

very nice! this is one of my fav new features of react!

client/src/components/AppSidebar.tsx
```diff
@@ -38,3 +45,5 @@ export default class AppSidebar extends React.Component {
     );
   }
 }
+
+export default withTranslation("en-gb")(AppSidebar);
```
jnu: for my info, what is the `en-gb` doing here? and how does this relate to the `i18next.ts` below where we're setting up to detect locale from the browser?

client/src/i18n/i18next.ts
```diff
@@ -0,0 +1,32 @@
...
+  // learn more: https://github.com/i18next/i18next-http-backend
+  .use(HttpBackendApi)
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    lng: "en",
+    fallbackLng: "en",
```
jnu, line 18: should this be `en-gb` since that's the only one we've defined so far?

jnu (approved):
looks good!!

I've never used this library before tho and i just want to confirm that we are definitely auto-detecting locale from the browser, and loading the correct file (if available).

client/src/App.tsx
seerocode: I didn't know what this was until an error told me to add it in. The loading element is a placeholder though and we should add a real one in.

client/src/components/AppSidebar.tsx
seerocode: This is supposed to specify the use of the en-gb translations we have in the locales folder. So if the local detected is en-gb, it will use these translations. But tests are necessary to make sure it works because UI-wise, it's not. :/

seerocode: <strike>Oh, I think I was using the wrong thing! Should probably be `withNamespaces()` instead: https://react.i18next.com/legacy-v9/step-by-step-guide</strike>

NVM!

client/src/i18n/i18next.ts
seerocode: I can change it!

seerocode:
Tests added!

client/src/__tests__/AppSidebar.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+            "teamsSideBarTitle": "My Programs"
+          }
+        },
+        lowerCaseLng: true,
+      });
+  })
+
+  test('translates sidebar title text to UK English', () => {
+    const sidebar = shallow(<AppSidebar />);
+    expect(sidebar.find(SubMenu).first().prop('title')).toBe('My Programmes');
```
jnu, line 37: 🤗 😍 yay tests

client/src/i18n/i18next.ts
jnu: this looks fine to me now -- were you able to figure out how to change the language? was it some cookie issue?

jnu (approved):
looks great to me!

client/src/i18n/i18next.ts
seerocode: @jnu Just pushed a change to correct it! The `lng` key in the i18next setup overrides language detection so I removed it.

PR #13, opened 2021-03-15
Ch32/graphql schema

nmanduj1:


api/mutations.py
```diff
@@ -1,11 +1,11 @@
 from ariadne import convert_kwargs_to_snake_case, ObjectType
 
 mutation = ObjectType("Mutation")
-@mutation.field("createUser")
+@mutation.field("upsertUser")
```
jnu, line 4: just a style note, decorators are usually written as a stack directly on top of the function they modify, with no lines in between, like:

```py
@mutation.field("createUser")
@convery_kwargs_to_snake_case
def resolve_upsert_user(...):
    ...
```

this more clearly highlights what the decorator is wrapping.

api/queries.py
```diff
@@ -1,17 +1,49 @@
...
+        'first_name': 'Cat',
+        'last_name': 'Berry',
     }
+
+    return payload
+
+@query.field("teamUsers")
+@convert_kwargs_to_snake_case
+
+def resolve_teamUsers(obj, info, team_id):
```
jnu: is this camelcase just bc of ariadne?

api/queries.py
```diff
@@ -1,17 +1,49 @@
 from ariadne import convert_kwargs_to_snake_case, ObjectType
 
 query = ObjectType("Query")
-@query.field("users")
 
+@query.field("user")
 @convert_kwargs_to_snake_case
 
-def resolve_users(obj, info):
+def resolve_user(obj, info, user_id):
```
jnu: I know part of the idea with FastAPI is to add typing info to functions. Is this something we're planning on using? I have mixed feelings about types in python, so I could go either way -- just wanna make sure we're being consistent.

I'm a huge fan of adding documentation of arguments for most functions!! typical python documentation looks like:

```py
def resolve_user(obj, info, user_id):
    '''GraphQL query to find a user based on user ID.

    :param obj: Describe what obj is
    :param info: Describe what info is
    :param user_id: The user's ID
    :returns: Whatever payload is
    '''
    ...
```

this is called "restructured text" style docstrings - a lot of IDEs (like PyCharm) are able to parse the params and return arguments from that string and show helpful info. some more info here: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

api/schema.graphql
```diff
@@ -3,35 +3,114 @@ schema {
...
 }
 
 type Query {
-    users: UsersResult!
-    user(userId: ID!): UserResult!
+    # Retrieve a single user.
+    user(userId: ID!) : User!
+
+    # Retrieve all users associated with a specific team.
+    teamUsers(teamId: ID!) : [User]!
```
jnu: Rather than having specific `teamUsers` and `programUsers` queries, the more natural thing in GraphQL is to define a query in terms of a specific _node_ in your data graph (like, a "team" or a "program" or a "user"). these nodes have natural relationships with other nodes in the graph -- like a team has users, and a program has users, etc., which are defined by the _fields_ of their respective nodes.

so for example if we have the types:

```graphql
type Team {
  id: ID!
  users: [User]!
  # etc
}

type Program {
  id: ID!
  users: [User]!
  # etc
}

type User {
  id: ID!
  # etc
```

then we can define our queries in terms of these nodes in the graph:

```graphql
type Query {
  team(id: ID!) : Team
  program(id: ID!) : Program
  user(id: ID!) : User
}
```

and then to query, we can do:
```graphql
# instead of having a predefined `teamUsers` query:
query {
  team(id: "t123") {
    users {
      name
    }
  }
}

# instead of having a predefined `programUsers` query:
query {
  program(id: "p987") {
    users {
      name
    }
  }
}
```

this will make the schema a bit more powerful, because it lets the client choose what relationships they want to query, rather than trying to anticipate up front which queries the client will need (as you would do in a REST API).

---

for what it's worth, i'm also going to add there is a next level of this, which **we do not need to do here**, but you will encounter it some places (specifically with things based on Relay). this is to unite all the types with a Node interface and provide one single _node_ query to fetch _any object_, regardless of type. like so:
```graphql
interface Node {
  id: ID!
}

type User implements Node {
  id: ID!
  name: String!
}

type Team implements Node {
  id: ID!
  users: [User]!
}

type Query {
  node(id: ID!): Node
}
```
then you can query:
```graphql
query {
  node(id: "program123") { # or "team987" or "user142" or whatever
    ... on Team {
      user { name }
    }
    ... on Program {
      user { name }
    }
    ... on User {
      name
    }
  }
}
```
the advantage of this is it can simplify keeping objects on the client in sync with the server very easily -- you can always fetch the latest data with the `node` query just by knowing the object's ID, regardless of what the object is.

api/schema.graphql
```diff
@@ -3,35 +3,114 @@ schema {
...
-    errors: [String]
+type DeleteUserOutput {
+    id: ID!
 }
 
 type Query {
-    users: UsersResult!
-    user(userId: ID!): UserResult!
+    # Retrieve a single user.
+    user(userId: ID!) : User!
```
jnu: should the return type be optional (if the user was not found)? or are we just anticipating an error here

api/schema.graphql
```diff
@@ -3,35 +3,114 @@ schema {
...
+
+    # Retrieve all users associated with a specific program.
+    programUsers(programId: ID!) : [User]!
 }
 
 type Mutation {
-    createUser(firstName: String!, lastName: String!): UserResult!
-    deleteUser(userId: ID!): DeleteUserResult!
+    # Upsert a user.
+    upsertUser(input: UpsertUserInput) : UpsertUserOutput!
```
jnu: just curious why we're using an upsert here? i think it's fine, but just need to be careful about which fields are mutable and which aren't! -- that's a future thing to keep in mind.

api/schema.graphql
```diff
@@ -3,35 +3,114 @@ schema {
...
+
+    # Delete a user.
+    deleteUser(input: DeleteUserInput!) : DeleteUserOutput!
```
jnu: i think this can just be an `userId: ID!` of the user for input, right?

api/schema.graphql
```diff
@@ -3,35 +3,114 @@ schema {
...
-    users: [User]
+input DeleteUserInput {
+    id: ID!
 }
 
-type DeleteUserResult {
-    success: Boolean!
-    errors: [String]
+type DeleteUserOutput {
+    id: ID!
```
jnu: this is fine, but also don't feel compelled to return anything if nothing makes sense -- it'd be reasonable for Deletes to return nothing (or an optional error).

jnu (changes requested):
great start! i left some feedback on how to structure the graphql schema which i think is important -- i also left some smol notes about style / documentation practices (which we ought to have documented elsewhere 😖)

api/schema.graphql
nmanduj1: Thanks Joe! This makes a lot of sense. I will update!

api/schema.graphql
nmanduj1: Yes! I meant to bring this up to the team. Should we be aiming for consistency in how the mutations are structured? aka Should all of our mutations use the input/ output pattern? Or would we prefer breaking from that pattern where it would be overkill? I figured just following the pattern for consistency but see merit to breaking when it makes sense.

api/schema.graphql
nmanduj1: Good catch!

api/queries.py
nmanduj1: Thanks Joe! I'll add this documentation to the resolvers-

api/schema.graphql
nmanduj1: hm, I wonder if graphql gives you the option of specifying which field are mutable.. i will look into this today.

api/queries.py
nmanduj1: yup, changing it to underscore gives me the following error: 'ValueError: Field upsert_user is not defined on type Mutation'
I left it as is just to get it running but I agree it should be underscored. I expect ariadne provides some way to handle the case mapping on '@query.field' but it's not something ive found in their docs yet. I will do some digging today.

nmanduj1:
@jnu, i made the changes you requested- would you be able to take another look?

api/README.md
```diff
@@ -56,13 +56,36 @@ Navigate to http://localhost:8000/graphql/ to see your GraphQL Playground in act
...
 ```
+
+sample mutation using inputs:
+
+```
+mutation upsertUsers ($userInfo : UpsertUserInput) {
+  upsertUser (input: $userInfo) {
+      user {
+        firstName
+      	lastName
```
jnu: nit: that alignment 😰 😰

api/README.md
```diff
@@ -56,13 +56,36 @@ Navigate to http://localhost:8000/graphql/ to see your GraphQL Playground in act
...
+      }   
+    }
+  }
+```
+
+in the playground's query variables section:
+```
+ {
+      "userInfo": {
```
jnu: glad u got this working!!

api/schema.graphql
```diff
@@ -3,35 +3,110 @@ schema {
     mutation: Mutation
 }
 
+scalar Date
+
+type Organization {
+    id: ID!
+    name: String!
+    teams: [Team]
```
jnu: graphql's "non-nullable" flag (the exclamation point) is a little tricky with lists. there are 4 permutations that mean slightly different things:

```graphql
# 1. Optional list, optional inner type:
# Valid values: null, [], [null], [MyTeamObj], [null, MyTeamObj]
# Invalid values: (nothing)
teams: [Team]

# 2. Optional list, required inner type:
# Valid values: null, [], [MyTeamObj]
# Invalid values: [null], [null, MyTeamObj]
team: [Team!]

# 3. Required list, optional inner type
# Valid values: [], [null], [null, MyTeamObj]
# Invalid values: null
team: [Team]!

# 4. Required list, required inner type
# Valid values: [], [MyTeamObj]
# Invalid values: null, [null], [null, MyTeamObj]
team: [Team!]!
```

In most (if not all!) of this schema I think you _do_ want at least the inner type to be required (so, `Team!` here). I personally find it easier to use option 4 where the list and the inner values are required (`[Team!]!`) everywhere, because there's less ambiguity. You just use an empty list to signify "no data" and you don't have to write checks in code for whether the list exists or not before you use it. I'll let you and @seerocode decide what you prefer though!

api/schema.graphql
jnu: seems like something you'd have to handle in python, in your resolver. I would do something like `raise ValueError('field not mutable')` if trying to update a non-mutable field. totally fine to do this in a future PR!

api/schema.graphql
jnu: the guy who created python actually uses a quote i agree with to talk about this, from ralph waldo emerson:

```
A foolish consistency is the hobgoblin of little minds
```

the point being, in many cases consistency is good -- formatting the code consistently makes it easier to read, which is really important. but if you're only trying to follow a pattern for the sake of the pattern, and the pattern is actually making things _less_ convenient, in that case you should break the pattern!

jnu (changes requested):
this looks great!!! the non-nullable issue is important to sort out with @seerocode before merging, since it'll affect how the typescript code interacts with the API. otherwise everything looks awesome, nice work!

api/schema.graphql
nmanduj1: excited team and excited array it is!

nmanduj1:
@jnu, i made the latest changes we talked about- feel free to review again if you'd like but I think we're good to merge.

jnu (approved):
it looks great!! love to see the excited arrays!!!!

PR #14, opened 2021-03-15
[ch84] Accessibility automated tests

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/84

### Description:
- Added [Eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) for a11y linting checks
- Added [react-axe](https://github.com/dequelabs/react-axe) to view a11y violations in console for pre-production environments 
- Added [jest-axe](https://github.com/nickcolley/jest-axe) plugin for Jest testing to extend tests for a11y

#### Still to consider that can't be caught by automatic testing:
- Principles of WCAG: https://www.w3.org/WAI/WCAG21/Understanding/intro#understanding-the-four-principles-of-accessibility
- Contrast ratio checks
- Screen reader checks


seerocode:
The test to check for accessibility violations looks like this:

```
describe('accessibility', () => {
    test('should not have basic accessibility issues', async () => {
      // TODO: Update with enzyme when released for React 17 (https://github.com/enzymejs/enzyme/issues/2429)
      // const sidebar = mount(<AppSidebar />);
      // const results = await axe(sidebar.getDOMNode()); 

      const { container } = render(<AppSidebar />)
      const results = await axe(container)
      expect(results).toHaveNoViolations();
    });
  });
```

I'll include tests in another PR since it's quite a bit to fix 😬 
Also wondering if we should separate accessibility tests differently? In a way we can run them in our pipeline separately?

client/.eslintrc.js
```diff
@@ -15,7 +15,9 @@ module.exports = {
   extends: [
     "plugin:react/recommended", // Uses the recommended rules from @eslint-plugin-react
     "plugin:@typescript-eslint/recommended", // Uses the recommended rules from @typescript-eslint/eslint-plugin
+    "plugin:jsx-a11y/recommended"
```
seerocode, line 18: Do we want strict or recommended? See documentation: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#supported-rules

jnu: recommended seems fine!

client/src/index.tsx
```diff
@@ -6,12 +6,24 @@ import "./i18n/i18next";
...
-ReactDOM.render(
-  <BrowserRouter>
-    <App />
-  </BrowserRouter>,
-  document.getElementById("root")
-);
+if (process.env.NODE_ENV !== 'production') {
+  import('react-axe').then(axe => {
+    axe.default(React, ReactDOM, 1000);
+    ReactDOM.render(
```
jnu, line 33: for my info, does the `ReactDOM.render` call have to come after the `axe.default` call for some reason? or are u trying to avoid some re-rendering from the async loading? it's fine either way, but i'd suggest moving the render lines into a function you can call, so that you don't have to duplicate the code in the `if` and in the `else`, like:

```js
function renderApp() {
  ReactDOM.render(
    <BrowserRouter>
      <App />
    </BrowserRouter>,
    document.getElementyById('root')
  );
}
```
and

jnu (approved):
looks great, nice work!!!

> Also wondering if we should separate accessibility tests differently? In a way we can run them in our pipeline separately?

do these tests take a while to run? IMO it's only worth breaking them out separately if they're relatively slow or unwieldy -- otherwise might as well just run them all the time

client/src/index.tsx
seerocode: @jnu I copied from the docs but your suggestion is what I thought of when I saw the duplication. I'll change it up now.

seerocode: Done! I'll go with my instincts next time. Also @jnu , here are the axe violations so far: 
image

seerocode:
Violations will be resolved in a separate PR.

PR #15, opened 2021-03-15
[ch54] Connect FE dev app to API

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/54

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -124,5 +124,6 @@
     "ts-jest": "^26.5.3",
     "tslint": "^6.1.3",
     "tslint-plugin-prettier": "^2.3.0"
-  }
+  },
+  "proxy": "http://localhost:8000/graphql/"
 }
--- a/client/package.json
+++ b/client/package.json
@@ -125,5 +125,5 @@
     "tslint": "^6.1.3",
     "tslint-plugin-prettier": "^2.3.0"
   },
-  "proxy": "http://localhost:8000/graphql/"
+  "proxy": "http://localhost:8000"
 }


client/package.json
```diff
@@ -124,5 +124,6 @@
     "ts-jest": "^26.5.3",
     "tslint": "^6.1.3",
     "tslint-plugin-prettier": "^2.3.0"
-  }
-}
+  },
+  "proxy": "http://localhost:8000"
```
nmanduj1, line 128: Hm, I'd expect this info to be in the apollo configuration/initializer, not in package.json. But I'm not sure how the proxy is being used here. Are you able to hit this endpoint successfully?

nmanduj1: I just read the docs and it looks like you are completely on point- this the correct implementation for this proxy. I usually just add a flag to route traffic depending on whether the node_env is set to development. Do you know what the pros are to setting this up this way?

seerocode: @nmanduj1 yep! This is being brought in with the Apollo Client in a different PR.

seerocode: @nmanduj1 From the docs I understood this is for development and avoids having to change your host everywhere when you deploy to production. There is [middleware](https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually) that provides additional flexibility should we need it or we can set up env files. This seemed to be a good start although I considered the middleware.

nmanduj1: thank u!

nmanduj1 (approved):

PR #16, opened 2021-03-15
[ch87] Add Apollo client

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/87


client/src/cache.ts
```diff
@@ -0,0 +1,33 @@
...
+export const cache: InMemoryCache = new InMemoryCache({
+    typePolicies: {
+        Query: {
+            fields: {
+                launches: {
+                    keyArgs: false,
+                    // This merge function takes our existing cached launches and 
+                    // the incoming launches and combines them into a single list, 
+                    // which it then returns. The cache stores this combined list 
+                    // and returns it to all queries that use the launches field.
```
jnu: for my info, where is this coming from? my general philosophy is to avoid caching until it's really necessary for performance reasons - just causes lots of headaches when developing. an in-memory cache might be ok, but im a little hesitant!

jnu (approved):
looks great! i'd say probably remove caching for now unless there's a compelling reason to have it :)

client/src/cache.ts
seerocode: @jnu We can delete this function and have an inmemory cache. At least from the docs, cache is a required field for the [ApolloClient](https://github.com/stanford-policylab/bbc-50-50/pull/16/files#diff-e1037d6d8abe24e1273b402e4ed0ec490b5d522c0d267c507de7d5fa705f28bfR16) instantiation.

jnu:
ok then that seems fine. thanks!

пн, 15 мар. 2021 г. в 18:09, Ro E ***@***.***>:

PR #18, opened 2021-03-16
Add workflows to run py / ts tests

jnu:
Set up tests to run on pull requests and merges into main

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,45 @@
+name: Python
+
+on:
+  push:
+    branches: [main]
+    paths: [api/**]
+  pull_request:
+    branches: [main]
+    paths: [api/**]
+
+jobs:
+  build:
+    name: Test Python
+    runs-on: ubuntu-latest
+    env:
+      wd: ./api
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+        id: python
+
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.head_ref}}
+
+      - name: Cache requirements
+        id: cache-reqs
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      - name: Get requirements
+        working-directory: ${{env.wd}}
+        run: |
+          python -m pip install -r requirements.txt
+
+      - name: Test
+        working-directory: ${{env.wd}}
+        run: python -m unittest -v
--- a/.github/workflows/ts.yml
+++ b/.github/workflows/ts.yml
@@ -27,9 +27,7 @@ jobs:
         working-directory: ${{env.wd}}
 
       - name: Configure Node.js
-        uses: actions/setup-node@v1.4.2
-        env:
-          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+        uses: actions/setup-node@v2
         with:
           node-version: "14"
 
--- a/.github/workflows/ts.yml
+++ b/.github/workflows/ts.yml
@@ -1,0 +1,47 @@
+name: Typescript
+
+on:
+  push:
+    branches: [main]
+    paths: [client/**]
+  pull_request:
+    branches: [main]
+    paths: [client/**]
+
+jobs:
+  build:
+    name: Build App
+    runs-on: ubuntu-latest
+    env:
+      wd: ./client
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


jnu:
fixes https://app.clubhouse.io/stanford-computational-policy-lab/story/85/set-up-ci-testing

.github/workflows/python.yml
```diff
@@ -0,0 +1,45 @@
+name: Python
+
+on:
```
seerocode, line 3: From the docs, this runs the workflow on push to any branch and when a PR is opened. How does it work for push to branch? Would you get some error on push if tests fail?

.github/workflows/python.yml
```diff
@@ -0,0 +1,45 @@
...
+  push:
+    branches: [main]
+    paths: [api/**]
+  pull_request:
```
seerocode, line 7: Could push and pull_request be combined in a list like [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events) or is it better to keep them separate?

.github/workflows/python.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+    name: Test Python
+    runs-on: ubuntu-latest
+    env:
+      wd: ./api
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
seerocode, line 19: <strike>Are these built in actions?</strike> NVM! Looked them up.

.github/workflows/python.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          ref: ${{github.head_ref}}
+
+      - name: Cache requirements
+        id: cache-reqs
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
```
seerocode, line 35: Looking these [keys](https://github.com/actions/cache) up might have answered my previous question! Does this make it so the workflow caches requirements so they aren't installed for each run? If so, what happens if the requirements are updated?

.github/workflows/ts.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Fetch merge base
+        if: ${{github.event_name == 'pull_request'}}
+        run: git fetch origin ${{github.base_ref}}
+        working-directory: ${{env.wd}}
+
+      - name: Configure Node.js
+        uses: actions/setup-node@v1.4.2
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
```
seerocode: Last question! Why is this required to be true?

.github/workflows/python.yml
jnu: not quite, the `push` hook is filtered to `main` -- so the tests will only run when we push to the main branch. This happens when we 1) merge pull requests and 2) the rare occasion when someone pushes directly to main. No other branch pushes will trigger the tests.

it's good to have the build run whenever the main branch is touched just to be sure, sometimes things can go wrong when you merge and even when your tests passed in a feature branch they will break in main.

.github/workflows/python.yml
jnu: do u mean the style where they have `on: [push, pull_request]`? or something different?

we can't use that inline style, because we want to have these filters set up (e.g., we only want to test python when code in the `api/` directory changes).

.github/workflows/python.yml
jnu: yea,, actually i just copied this from one of our other projects where it's more relevant. The cache is keyed by the hash of `requirements.txt`, so when the requirements.txt changes in any way it'll invalidate the cache and we'll have to redownload.

this speeds up the PR builds a bit. the project I copied it from it makes much more of a difference, since there we have a 1GB NLP model to download, so without caching it'd be pretty slow.

.github/workflows/ts.yml
jnu: ah good question, i just copied this again from a different project.

it was related to this: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

but it looks like the setup-node action has since been patched fully, so i can try removing it

.github/workflows/python.yml
seerocode: Thanks for explaining @jnu !

.github/workflows/ts.yml
seerocode: Sounds good! Do you have to specify a version for this patch?

.github/workflows/python.yml
seerocode: Ohh, okay!

.github/workflows/ts.yml
jnu: yep! just bumped to version 2, we'll see how that goes

seerocode (approved):
🚢

PR #19, opened 2021-03-16
Ability to auto generate types

jnu:
Yarn commands to introspect the graphql endpoint.

When you write gql queries on the frontend, you can run the `yarn run gentypes` command to generate typings for the response.

The `yarn run introspect` command will download the schema as `schema.json`.

You will need the Python API running to run either of these commands

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -121,6 +121,7 @@
     "@types/enzyme-adapter-react-16": "^1.0.6",
     "@types/jest": "^26.0.20",
     "@types/jest-axe": "^3.5.0",
+    "apollo": "^2.32.5",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.6",
     "enzyme-to-json": "^3.6.1",
--- a/client/package.json
+++ b/client/package.json
@@ -31,7 +31,9 @@
     "test:watch": "jest --watch",
     "eject": "react-scripts eject",
     "lint": "tslint -c tslint.json 'src/**/*{.ts,.tsx}'",
-    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'"
+    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'",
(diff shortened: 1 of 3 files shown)


client/package.json
```diff
@@ -31,7 +31,9 @@
     "test:watch": "jest --watch",
     "eject": "react-scripts eject",
     "lint": "tslint -c tslint.json 'src/**/*{.ts,.tsx}'",
-    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'"
+    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'",
+    "gentypes": "apollo client:codegen --target typescript --endpoint=http://localhost:8000/graphql --excludes='node_modules/*' --outputFlat='src/__generated__' --includes='src/**/*.ts*' --useReadOnlyTypes && pretty-quick --pattern 'client/src/__generated__/**/*.ts' --branch HEAD",
+    "introspect": "apollo schema:download src/schema.json --endpoint=http://localhost:8000/graphql"
```
seerocode, line 36: So cool! This worked for me.

client/package.json
```diff
@@ -31,7 +31,9 @@
     "test:watch": "jest --watch",
     "eject": "react-scripts eject",
     "lint": "tslint -c tslint.json 'src/**/*{.ts,.tsx}'",
-    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'"
+    "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'",
+    "gentypes": "apollo client:codegen --target typescript --endpoint=http://localhost:8000/graphql --excludes='node_modules/*' --outputFlat='src/__generated__' --includes='src/**/*.ts*' --useReadOnlyTypes && pretty-quick --pattern 'client/src/__generated__/**/*.ts' --branch HEAD",
```
seerocode, line 35: I had to yarn install "apollo" for this to work. Not sure why since Apollo is in this project. What happens when you run this command on your end?

jnu: ah, good catch - i just added the `apollo` package which includes the CLI (we only had the JS client so far) as a dev dependency, so it should work now

seerocode (approved):
🚢

nmanduj1 (approved):

PR #21, opened 2021-03-29
Ch27/users management

nmanduj1:


api/app.py
```diff
@@ -1,14 +1,67 @@
...
+
+class UserUpdate(User, models.BaseUserUpdate):
+    pass
+
+
+class UserDB(User, models.BaseUserDB):
+    pass
+
+
+DATABASE_URL = "postgres://postgres:pass@localhost/bbc"
```
jnu: looks like the way to set up a config with users and passwords and stuff in FastAPI is pretty easy so you can add it here. it uses the feature called `Settings`: https://fastapi.tiangolo.com/advanced/settings/#pydantic-settings

you can add a `Settings` field here in app.py for now, we'll probably want to move it later. I'd say all of the user, password, host, and database should be config settings, since all four are likely to change between deployments. you can keep some defaults if you want, like:

```py
class Settings(BaseSettings):
    db_user: str = 'postgres'
    db_pw: str = 'pass'
    db_host: str = 'localhost'
    db_name: str = 'bbc'
```

api/app.py
```diff
@@ -1,14 +1,67 @@
...
+
+database = databases.Database(DATABASE_URL)
+
+Base: DeclarativeMeta = declarative_base()
+
+
+class UserTable(Base, SQLAlchemyBaseUserTable):
+    pass
+
+engine = sqlalchemy.create_engine('postgresql+psycopg2://postgres:pass@localhost/bbc')
```
jnu: why are we creating a second database connection here? (or, how does this differ from L36 where we're using `DATABASE_URL`?

api/requirements.txt
```diff
@@ -12,3 +12,7 @@ python-dotenv==0.15.0
 six==1.15.0
 SQLAlchemy==1.3.23
 Werkzeug==1.0.1
+fastapi-sqlalchemy==0.2.1
+# fastapi_users==5.1.2
```
jnu, line 17: do we not need this fastapi_users, or why is it commented out?

api/app.py
```diff
@@ -1,14 +1,67 @@
...
+
+
+users = UserTable.__table__
+user_db = SQLAlchemyUserDatabase(UserDB, database, users)
+
+
 app = FastAPI()
 
+# auth_backends = [
+#     JWTAuthentication(secret=SECRET, lifetime_seconds=3600),
```
jnu: a JWT token sounds great! is this all not quite ready tho?

jnu (changes requested):
woohoo! great start!! i don't have full context tho so maybe tomorrow morning let's walk through it together?

jnu (approved):

PR #22, opened 2021-03-29
Add client-side mock graphql server

seerocode:
### Description

A mock GraphQL apollo server on client-side to use for feature testing.

Did we need it? No.
Is it perfect? Not even close.
Did I do it anyway? Yep!


client/apollo.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    client: {
+        service: {
+            name: 'reporting-tool-service',
+        },
+        excludes: ['src/__mocks__/*']
```
seerocode, line 6: Without this, the gentypes script fails if there's a query inside this directory.

client/src/__mocks__/gql-mock-server.ts
```diff
@@ -0,0 +1,35 @@
+import { ApolloServer, gql } from "apollo-server";
+import { schema } from "./schema"
+import fake from "casual";
+
+const mockTypes = {
```
seerocode, line 8: Could be moved out to a mockQuery file as well.

client/src/__mocks__/schema.ts
```diff
@@ -0,0 +1,103 @@
+import { gql } from "apollo-server";
+
+const schema = gql`
```
seerocode: @jnu I haven't figure out yet how to bring this in from the client side so it's not a static schema. I figured the gentypes script would do it and I got stuck there.

client/src/__mocks__/gql-mock-server.ts
```diff
@@ -0,0 +1,31 @@
...
+  }),
+  Program: () => ({
+    name: "BBC News",
+  }),
+};
+
+const server = new ApolloServer({
+  typeDefs: schema,
+  mocks: mockTypes,
+  // mocks: true,
```
jnu: maybe remove the commented stuff if they're not needed?

client/src/__mocks__/schema.ts
jnu: there should be a file called `client/src/schema.json` already checked in to the repo. this is generated by running `yarn run introspect` -- can you use that here?

jnu (changes requested):
lgtm! but i agree we should figure out how to make sure we're not duplicating the schema

client/src/__mocks__/schema.ts
seerocode: Yeah I see it but I don't know how to bring it in. I tried following https://www.apollographql.com/docs/apollo-server/testing/mocking/#mocking-a-schema-using-introspection but the `buildClientSchema` function kept erroring for the json file.

jnu: ok ugh, i just looked into this. It looks like there must be a version mismatch and/or outdated documentation about this, but i got it working.

in the `schema.json` that we have, you don't need the `data` attribute, despite what it says. I also got some typescript errors that seemed irrelevant. the following works for me:

```ts
import { buildClientSchema } from 'graphql';
// NOTE: using the `require` format bypasses typescript's type-checking, which appears to have a
// slight incompatibility here. (You can look more into this if you want -- might need to upgrade a version
// somewhere or something).
const introspectedSchema = require('../schema.json');

// [mockTypes, etc]

const server = new ApolloServer({
  // NOTE: `introspectedSchema` is the `any` type here, but it works!
  schema: buildClientSchema(introspectedSchema),
  mocks: mockTypes,
  playground: true,
});
```

jnu: Let me know if this works for you -- and if so we should be able to drop the new `schema.ts`, which is my main concern about this pr! -- just don't want to duplicate the schema.

client/README.md
```diff
@@ -6,6 +6,10 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 
 In the project directory, you can run:
 
+### `yarn mock-server`
+
+Runs a mock GraphQL server on the client side. Open [http://localhost:4000](http://localhost:4000) to view the GQL playground in the browser.
```
jnu: Could you add a note here about running `yarn introspect` to update the schema.json? This must be done with the api running. (We could consider doing an automated check to update the schema.json automatically when `mock-server` is run, if the API is reachable, but i'd say leave it dumb for now and just add it to the readme.)

client/src/__mocks__/gql-mock-server.ts
```diff
@@ -0,0 +1,29 @@
+import { ApolloServer, gql } from "apollo-server";
+import { schema } from "./schema";
+import fake from "casual";
```
jnu, line 4: this is a cool library, i've never used it before!

when i'm playing with this library though i notice one of my fav programming "gotchas" to go over!

the topic is: _randomness_. we should be very thoughtful about using randomness. we should only use it in the small number of contexts where it's really required, such as security.

the reason is: especially in tests (but even in normal code), we want everything to be predictable and deterministic, so that we can make as strong a claim as possible that it's correct. there are legitimate classes of bugs that show up due to natural/unavoidable randomness (such as race conditions due to network latency). these bugs are hard enough to debug on their own (i have stories!!), and so we should control as much of our environment as we possibly can to keep ourselves sane.

it looks like `casual` has a seeding function you can call: `fake.seed(123)`. this will make sure that there's a deterministic pattern to the data it generates. i'm happy to talk more about this if you want -- there are some gotchas about seeding too!

client/src/__mocks__/gql-mock-server.ts
jnu: yep but this is fine for now!

jnu (changes requested):
this looks great!! i made a few small comments to address before merging, but this is cool and i think it'll be useful.

i'm really sorry that it took me so long to get back to this review!

seerocode:
Thank you! I'm excited to try out your changes.   
  
  
  
  
  
    
      
    

  

  
  
  
  
          
  

  
  
>

client/src/__mocks__/schema.ts
seerocode: @jnu I followed your code above and found that you can avoid the require import if you run `introspectedSchema as any` for `buildClientSchema`. It works now!

client/README.md
seerocode: @jnu I updated the documentation here and the script so the `introspect` script runs before the mock server starts up. Should bring in any latest changes that way.

client/package.json
```diff
@@ -33,7 +33,8 @@
     "lint": "tslint -c tslint.json 'src/**/*{.ts,.tsx}'",
     "lint:fix": "tslint --fix -c tslint.json 'src/**/*{.ts,.tsx}'",
     "gentypes": "apollo client:codegen --target typescript --endpoint=http://localhost:8000/graphql --excludes='node_modules/*' --outputFlat='src/__generated__' --includes='src/**/*.ts*' --useReadOnlyTypes && pretty-quick --pattern 'client/src/__generated__/**/*.ts' --branch HEAD",
-    "introspect": "apollo schema:download src/schema.json --endpoint=http://localhost:8000/graphql"
+    "introspect": "apollo schema:download src/schema.json --endpoint=http://localhost:8000/graphql",
+    "mock-server": "yarn introspect && ts-node-dev ./src/__mocks__/gql-mock-server.ts"
```
jnu: i think this is a good idea, but i would make the introspection step optional -- maybe you can do something like:

```
(curl --silent --fail localhost:8000/graphql && yarn introspect) ; ts-node-dev ./src/__mocks__/gql-mock-server.ts
```

this way it'll only introspect if the local server is available, and run the mock server either way.

client/README.md
```diff
@@ -1,11 +1,17 @@
...
 
 ## Available Scripts
 
 In the project directory, you can run:
 
+### `yarn mock-server`
+
+Downloads the GraphQL schema and mocks a GraphQL server on the client side. Open [http://localhost:4000](http://localhost:4000) to view the GQL playground in the browser.
+
+📝 **NOTE:** You must be running the API before running this command and pull in the latest changes from the main branch. See [API documentation here]("../../../api/README.md).
```
jnu: if we make running the API required, maybe it'd make sense to just put the mocks on the backend? (e.g., configure the API to startup with a mocked database instead of postgres?) i imagine it'll be useful to run the mock-server _without_ a hard dependency on the API (e.g., for front-end tests---setting up CI will be easier if we don't have to set up the python environment to run typescript tests).

my gut is also that we should remove the instruction to always run the API server from `main`---i think it's better to keep internal consistency within a feature branch, and then merge. (for example, maybe you've made schema changes in your feature branch that you want to mock, which are not yet in main.)

that all said, i think it's totally fine to automate the introspect step -- see my other comment about how to do this?

client/src/__mocks__/gql-mock-server.ts
```diff
@@ -0,0 +1,36 @@
...
+  Program: () => ({
+    name: "BBC News",
+  }),
+  Date: () => {
+    return fake.date();
+  },
+};
+
+const server = new ApolloServer({
+  schema: buildClientSchema(introspectedSchema as any),
```
jnu, line 32: yep, `as any` is fine! i get a little nervous about having the `any` type directly in the code, which is why i suggested using `require`, but in practice they do the same thing. (`require` just implicitly assigns the `any` type to whatever it imports.)

could you add a note in a comment to explain why we need the `any` type? it seems like some sort of version or documentation snafu in the apollo docs, and that it'll be fixed in the future. we could also report this to the apollo team on github, if you want!

jnu (approved):
thank you ro!!!

made a comment about whether we really wanna make the `introspect` step a strict requirement of the mock-server, but everything looks great. nice work! ✅

client/src/__mocks__/gql-mock-server.ts
jnu: some more context on this: https://github.com/apollographql/apollo-tooling/issues/1491

looks like the issue is that the typings are written in a slightly incorrect way, and state that certain _optional_ fields are actually _required_. This means that typescript can't infer that the (typeless) JSON file we have on hand is compatible with the type declaration expected by `buildClientSchema` (i.e., the `IntrospectionQuery` type).

it sounds like maybe the only way of getting the proper `IntrospectionQuery` type would be to execute the introspect query here in code, rather than on the CLI.

i don't think it's worth spending more time on! using `any` is totally fine here.

client/package.json
seerocode: I like this! I'll update to add that.

client/README.md
seerocode: That's fair. I'm going to use the command you suggested in a prev comment to avoid extra steps.

client/src/__mocks__/gql-mock-server.ts
seerocode: When I tried the require import, the linter threw an error even though I disabled eslint for that line so I looked for an alternative. Thanks for pointing to that issue!

client/src/__mocks__/gql-mock-server.ts
seerocode: This makes sense to me. I tried the seeding in the casual library and it doesn't seem to be sticking.

PR #23, opened 2021-03-29
[ch112] Landing page for Data Inputter

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/112

How she's lookin:
image

### DIFF
--- a/client/src/components/SearchAutoComplete.tsx
+++ b/client/src/components/SearchAutoComplete.tsx
@@ -17,7 +17,7 @@ const SearchAutoComplete = ({ dataSource }: Props) => {
       style={{ width: 300, float: "right", marginBottom: "10px" }}
       aria-label="Search"
       dropdownMatchSelectWidth={252}
-      options={dataSource}
+      dataSource={dataSource}
       placeholder={`Search your ${t("programs")}`}
       filterOption={(inputValue, option) =>
         option!.value.toUpperCase().indexOf(inputValue.toUpperCase()) !== -1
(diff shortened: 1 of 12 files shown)


client/src/components/SearchAutoComplete.tsx
```diff
@@ -0,0 +1,33 @@
...
+      style={{ width: 300, float: "right", marginBottom: "10px" }}
+      aria-label="Search"
+      dropdownMatchSelectWidth={252}
+      dataSource={dataSource}
+      placeholder={`Search your ${t("programs")}`}
+      filterOption={(inputValue, option) =>
+        option!.value.toUpperCase().indexOf(inputValue.toUpperCase()) !== -1
+      }
+      aria-expanded="false"
+      notFoundContent={`${capitalizeFirstLetter(t("program"))} not found`}
```
jnu: hm, i wonder if `toUpperCase` actually does the right thing for all languages / alphabets? not worth dwelling on, but maybe check first to see whether this i18n logic has any helper methods built in for cross-language capitalization

jnu (approved):
looks good! great work!! as for the tests, let's look at those together tomorrow and see if there's a way we can fix them without rewriting an entirely new component. thanks!!

client/src/components/SearchAutoComplete.tsx
seerocode: I ended up removing the function and replacing it with a css selector: https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter

PR #25, opened 2021-04-01
[Ch155] Update database schema and add mock data for datasets

seerocode:
- Updated base user table and db model to add first name and last name fields 
- Mostly changed the date time fields to timestamp (following the example in Nudge and [this good explainer](https://www.eversql.com/mysql-datetime-vs-timestamp-column-types-which-one-i-should-use/))


api/database.py
```diff
@@ -20,27 +22,31 @@ class Organization(Base):
...
     id = Column(Integer, primary_key=True)
-    name = Column(String(255))
+    name = Column(String(255), nullable=False)
     teams = relationship('Team')
 
-    created = Column(DateTime, nullable=False)
-    updated = Column(DateTime)
-    deleted = Column(DateTime)
+    created = Column(TIMESTAMP,
+                     server_default=func.now(), nullable=False)
```
jnu, line 34: perfect, thanks for adding these defaults!

api/mock_data.sql
```diff
@@ -0,0 +1,81 @@
...
+        'Berry',
+        1
+    );
+
+INSERT INTO
+    role
+VALUES
+    (
+        1,
+        'non-admin',
```
jnu: i would suggest for these roles that role-based access (RBAC) models usually operate on "positive" grants, not "negative" ones. what i mean by that is:

a role like `non-admin` is a "negative" role. in other words, it is specifying that a user does _not_ have a certain role (in this case, `admin`). in order to operate that system, we'd need to write logic that makes sure that 1) every user who is not an admin has the negative `non-admin` role tied to them, and 2) every admin user has the negative `non-admin` role removed and the positive `admin` role added.

a simpler system focuses only on adding positive roles, and making the negative case implicit. to make a user an `admin` we can assign them the positive `admin` role; everyone who does not have that role is implicitly a "non-admin".

this has two main advantages:
1) easier to write a query / build a UI to toggle the role, since it involves adding/removing a single field
2) supports a many:1 relationship of roles to users. this leaves us more flexibility in the future to expand our permission system to have more fine-grained roles, so that we don't have to put much up-front effort into guessing what roles we need right now.

jnu: a related thought here is that often i like to break up my db init scripts into two files:

1. data that is required for any deployment of the app. it'd be reasonable to define some roles in that file, like "admin".
2. dummy data for development -- i.e. most of everything else you put here

api/mock_data.sql
```diff
@@ -0,0 +1,81 @@
...
+        '2016-09-29 20:36:08'
+    );
+
+INSERT INTO
+    team
+VALUES
+    (1, 'News team', 1, now());
+
+INSERT INTO
+    "user"
```
jnu: i'd create at least one more user and assign them an admin role via the user_roles table (see my other comment :)

api/mock_data.sql
```diff
@@ -0,0 +1,81 @@
...
+VALUES
+    (
+        uuid_generate_v4(),
```
jnu: i made a comment in your other PR about randomness, and i think it applies here as well. it'll be a little easier to work in our dev environment if we hard-code an ID here instead of randomly generating it.

api/mock_data.sql
```diff
@@ -0,0 +1,81 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
jnu: i made another comment elsewhere about having an `init.sql` in addition to a `dev-data.sql` (or whatever you wanna call it). seems like this should be in the init.sql

api/database.py
```diff
@@ -54,12 +60,20 @@ class Team(Base):
 class User(Base, SQLAlchemyBaseUserTable):
```
jnu, line 65: i love how you added the `server_default` for the created/updated fields, i'm wondering if something similar is set up for the `user.id` column?

jnu (changes requested):
this is awesome and will be super helpful for development, thank you!!!

i made a couple of comments that are kind of wordy, but are actually very smol/quick things to change -- just wanna make sure you know why i'm suggesting them.

i also wanna make sure @nmanduj1 gets a chance to look over this if she hasn't already because i know she's actively working on these parts of the code!

api/database.py
seerocode: This is the statement sqlalchemy returns for the `Users` table:

```
CREATE TABLE "user" (
        id CHAR(36) NOT NULL, 
        email VARCHAR(320) NOT NULL, 
        hashed_password VARCHAR(72) NOT NULL, 
        is_active BOOLEAN NOT NULL, 
        is_superuser BOOLEAN NOT NULL, 
        is_verified BOOLEAN NOT NULL, 
        first_name VARCHAR(50) NOT NULL, 
        last_name VARCHAR(50) NOT NULL, 
        team_id INTEGER, 
        created TIMESTAMP DEFAULT now() NOT NULL, 
        updated TIMESTAMP DEFAULT now(), 
        deleted TIMESTAMP, 
        PRIMARY KEY (id), 
        FOREIGN KEY(team_id) REFERENCES team (id)
)
```

seerocode: This user.id column is supposed to be a UUID? @jnu

api/mock_data.sql
seerocode: Added an `init.sql` file for the no. 1 but not sure what else is needed there. Renamed to `init-dev.sql` for no. 2

api/database.py
```diff
@@ -203,3 +204,11 @@ class Record(Base):
     engine = create_engine('postgresql+psycopg2://' + DATABASE_URL)
 
     Base.metadata.create_all(engine)
+
+    init_roles = text(
+        "INSERT INTO role \
+            VALUES \
+            (1, '', 'User is a team member and has no administrative priveleges'), \
```
jnu: nit: `privileges` :)

api/database.py
```diff
@@ -168,12 +193,22 @@ class Record(Base):
...
+                     server_default=func.now(), onupdate=func.now())
+    deleted = Column(TIMESTAMP)
 
 
 if __name__ == '__main__':
     engine = create_engine('postgresql+psycopg2://' + DATABASE_URL)
 
     Base.metadata.create_all(engine)
+
+    init_roles = text(
```
jnu: since we have the db objects, do you want to avoid writing raw sql? like:
```
session = SessionLocal() # see nancy's pr
session.add(Role(name="admin", description="..."))
session.commit()
```

not a big deal either way, just a little more readable

api/database.py
```diff
@@ -168,12 +193,22 @@ class Record(Base):
...
+        "INSERT INTO role \
+            VALUES \
+            (1, '', 'User is a team member and has no administrative priveleges'), \
```
jnu: im not sure we need the empty string as a formal role, right? i forget what we landed with last week, but any role we create should definitely have a name. my instinct is to _not_ create any special role for normal users, because the implicitly have the default set of privileges. we only need to distinguish the _admin_ role currently, which provides a special set of permissions.

api/init-dev.sql
```diff
@@ -0,0 +1,62 @@
+INSERT INTO
+    organization
```
jnu: just wanna note it may be easier to make all these objects with the Python SqlAlchemy objects (like `Organization(name="BBC")` -- i'd imagine that would be a bit easier to work with, but it's up to you!

api/init.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+INSERT INTO
+    role
+VALUES
+    (
+        1,
+        '',
+        'User is a team member and has no administrative priveleges'
```
jnu: why do we need this in addition to the SQL you put in the python code? don't they do the same thing?

jnu (changes requested):
thanks ro!!! this is super closer, just made a few questions about how to clean up a little bit more

api/database.py
jnu: im not sure why it wouldn't be, but i wouldn't dwell too much on this, it shouldn't matter in practice -- the uuid type would be somewhat faster and smaller but it really shouldn't matter

api/database.py
seerocode: Thank you!!! This is what I needed.

api/database.py
seerocode: Fixed! I had misunderstood and it was apparent to me when I fixed this to be programmatic as mentioned in a previous comment.

api/init.sql
seerocode: Removed file.

api/init-dev.sql
seerocode: Deleted this mock data to move into the new story created (https://app.clubhouse.io/stanford-computational-policy-lab/story/194/set-up-mock-data-in-postgres)

jnu (approved):
this looks good to me,,, i think adding a script to populate with test data could be useful but you can definitely do it later!

PR #27, opened 2021-04-01
[ch124] Simple Dataset Details UI

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/124

image

### DIFF
--- a/client/src/components/DatasetRecordsTable.tsx
+++ b/client/src/components/DatasetRecordsTable.tsx
@@ -1,10 +1,9 @@
-import { Table, Typography } from "antd";
+import { Table } from "antd";
 import React from "react";
-import { ColumnsType } from "antd/es/table";
 import "./DatasetRecordsTable.css";
+import { Link } from "react-router-dom";
 
 const { Column, ColumnGroup } = Table;
-const { Text } = Typography;
 
 interface DatasetRecordsProps {
   datasetId: string;
@@ -39,7 +38,7 @@ type Disability = {
 interface Record {
   key: string;
   id: number;
-  airdate: string;
+  publicationDate: string;
   gender: Gender;
   race?: Race;
   ethnicity?: Ethnicity;
@@ -51,7 +50,7 @@ const data: Record[] = [
   {
     key: "1",
     id: 1,
-    airdate: "12/11/2020",
+    publicationDate: "12/11/2020",
     gender: {
       men: 1,
       women: 2,
@@ -68,7 +67,7 @@ const data: Record[] = [
   {
     key: "2",
     id: 2,
-    airdate: "12/12/2020",
+    publicationDate: "12/12/2020",
     gender: {
       men: 1,
       women: 2,
@@ -84,6 +83,16 @@ const data: Record[] = [
   },
 ];
 
+const getValueGetter = <T extends { [key: string]: number }>(key: keyof T) => {
+  return (record: T | undefined) => {
+    if (!record) {
+      return 0;
+    }
+
+    return record[key] || 0;
+  };
+};
+
 const DatasetRecordsTable = ({
   datasetId,
 }: DatasetRecordsProps): JSX.Element => {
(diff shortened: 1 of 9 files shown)


client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+            return <Text>{gender.nonBinary}</Text>;
+          }}
+        />
+      </ColumnGroup>
+
+      <ColumnGroup title="Race">
+        <Column
+          title="Black"
+          dataIndex="race"
+          render={(race: Race) => {
```
seerocode: There's some duplication with these rendered functions and I'm working on refactoring that.

seerocode:
Adding tests also!

client/src/components/DatasetRecordsTable.tsx
jnu: some duplication is ok -- it's not usually worth generalizing on the first pass -- but it's a good instinct and i'll see if i can make some suggestions!

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+            if (typeof race && typeof race.black !== "undefined") {
```
jnu: you've opened a big can o' worms in javascript ... _undefined_

`typeof race` always evaluates to true. if `race` was never defined, `typeof race === 'undefined'`. since `'undefined'` is a non-empty string, it will be interpreted as `true`.

`typeof race.black` might be `'undefined'`, but there are more idiomatic ways to check this. i'd suggest rewriting this function as something like:

```ts
(race: Race) => {
  if (!race || !race.hasOwnProperty('black') {
    return 0;
  }
  return race.black;
}
```

there's also some slight funniness with your function signature -- if you expect `race` might not exist, it should be marked as optional in the argument type. under what circumstances is `Race` undefined?

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+              return <Text>{race.black}</Text>;
+            }
+            return 0;
+          }}
+        />
+        <Column
+          title="White"
+          dataIndex="race"
+          render={(race: Race) => {
+            if (typeof race && typeof race.white !== "undefined") {
```
jnu: the pattern i'd use to generalize this is a factory function. this uses a closure to create a function that comes configured to use a given attribute:

```ts
const getRaceRenderer = (raceKey: keyof Race) => {
  return (race: Race?) => {
    if (!race) {
      return 0;
    }
    return race[raceKey] || 0;
  };
}
```

then down here you can use:
```ts
render={getRaceRenderer('white')}
```

let me know if you have questions or want to pair on this!! the `keyof` in particular might raise an eyebrow :)

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+};
+
+// TODO: populate type based on dataset options for race
+type Race = {
+  black?: number;
+  white?: number;
+  asian?: number;
+};
+
+// TODO: populate type based on dataset options for ethnicity
```
jnu, line 25: hehe, this lil `TODO` might turn into a big headache :)

let's circle back to this later, though -- i think this is a great start for now

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+  key: string;
+  id: number;
+  airdate: string;
+  gender: Gender;
+  race?: Race;
+  ethnicity?: Ethnicity;
+  disability?: Disability;
+}
+
+// TODO: add gql query to get dataset id records based on user
```
jnu, line 48: is it worth merging your mock server pull request and setting up the GQL query now, in this pull request?

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+              typeof ethnicity &&
+              typeof ethnicity.hispanicLatinx !== "undefined"
+            ) {
+              return <Text>{ethnicity.hispanicLatinx}</Text>;
+            }
+            return 0;
+          }}
+        />
+        <Column
+          title="Western European"
```
jnu, line 154: as you mentioned in your TODO above, in the future we'll want to generate these columns dynamically, but this is fine for now!

jnu (approved):
this looks great to me! i made a couple of small comments about refactoring and types, but im going to pre-emptively approve to unblock you. let me know if you want to go over anything!

client/src/components/DatasetRecordsTable.tsx
seerocode: @jnu thank you!!!! I wasn't sure how to pass a generic type and `keyof` is what I was looking for.

EDIT: I read this a little wrong. I had something similar to this but I wanted to know how to pass a generic Type that could be `Race` or `Ethnicity`, etc.

PR #29, opened 2021-04-06
[ch129] Data entry UI

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/129

#### Screenshots
image

### DIFF
--- a/client/src/components/AggregateDataEntryForm.tsx
+++ b/client/src/components/AggregateDataEntryForm.tsx
@@ -1,0 +1,85 @@
+import React, { useState } from "react";
+import { Card } from "antd";
+import "./AggregateDataEntryForm.css";
+
+const AggregateDataEntryForm = (): JSX.Element => {
+  const handleFocus = (event: { target: { value: string } }) => {
+    event.target.value = "";
+  };
+
+  //TODO: set state for each entry
+
+  return (
+    <Card type="inner" title="Gender">
+      <form>
+        <label htmlFor="women">
+          <input
+            type="number"
+            id="women"
+            className="data-input-form_field"
+            defaultValue="0"
+            onFocus={handleFocus}
+          />{" "}
+          Women
+        </label>
+
+        <label htmlFor="men">
+          <input
+            type="number"
+            id="men"
+            className="data-input-form_field"
+            defaultValue="0"
+            onFocus={handleFocus}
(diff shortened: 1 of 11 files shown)


client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useState } from "react";
+import { Card } from "antd";
+import "./AggregateDataEntryForm.css";
+
+const AggregateDataEntryForm = (): JSX.Element => {
+  const handleFocus = (event: { target: { value: string } }) => {
+    event.target.value = "";
```
jnu: fine for "add data" but for edit probably don't want this.

maybe logic should be "if value == 0 then clear else leave it"

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <form>
+        <label htmlFor="women">
+          <input
+            type="number"
+            id="women"
+            className="data-input-form_field"
+            defaultValue="0"
+            onFocus={handleFocus}
+          />{" "}
+          Women
```
jnu, line 25: we've talked about this already but just flagging that in the future we'll want to generate all these inputs dynamically

client/src/components/DataEntry.tsx
```diff
@@ -0,0 +1,111 @@
...
+              <h4>Publication date:</h4>
+            </label>
+            <input type="date" id="publicationDate" name="publicationDate" />
+          </form>
+        </Col>
+      </Row>
+      <Row gutter={[16, 16]} className="data-entry">
+        <Col span={8}>
+          <h3 className="data-entry_category-descr-header">About gender</h3>
+          <Text>{`Gender identity expresses one's innermost concept of self as male,
```
jnu, line 66: i suppose this should technically come from a translation. we talked about letting the admin configure the description, but even then i think it's good to provide a default in the translation file.

jnu (approved):
looks great!! made a couple of small comments that don't need to be addressed here

client/src/components/AggregateDataEntryForm.tsx
seerocode: Done!

PR #32, opened 2021-04-12
Ch133/data entry mutation

nmanduj1:
Add create_dataset resolver logic

### DIFF
--- a/api/mutations.py
+++ b/api/mutations.py
@@ -46,7 +46,6 @@ def resolve_create_dataset(obj, info, input):
         :param id: Params to be changed 
         :returns: Newly created Dataset dictionary with eager-loaded associated Tags
     '''
-    #TODO Flush vs Commits??
     #TODO check docs- more efficient syntax to handle loading associated programs?
     #TODO check docs- is there a fancy decorator to handle automatic snakecase mapping of inputs?
 
@@ -77,6 +76,5 @@ def resolve_create_dataset(obj, info, input):
     session.commit()
 
     persisted_dataset = session.query(Dataset).filter(Dataset.id == dataset.id).options(joinedload("tags")).first().__dict__
-    persisted_dataset["tags"] = map(lambda tag: tag.__dict__, persisted_dataset["tags"])
-
+    persisted_dataset["tags"] = [tag.__dict__ for tag in persisted_dataset['tags']]
     return persisted_dataset
--- a/api/mutations.py
+++ b/api/mutations.py
@@ -1,5 +1,8 @@
 from ariadne import convert_kwargs_to_snake_case, ObjectType
+from settings import settings
+from database import SessionLocal, User
 
+session = SessionLocal()
 mutation = ObjectType("Mutation")
 
 @convert_kwargs_to_snake_case
@@ -39,11 +42,22 @@ def resolve_create_dataset(obj, info, input):
         :param obj: obj is a value returned by a parent resolver
         :param info: Has context attribute that contains ContextValue specific to the server implementation.
         :param input: Dataset to be created
-        :returns: 
-    '''
+        :returns: Created Dataset
+    # '''
+
+    # ins = session.query('dataset').insert().values(
+    #   name = 'Karan', 
+    #   description = "Jeans", 
+    #   programId = 1 
+    # )
+
+    cat = session.query(User)
+    print(f'{cat} love')
+    
     payload = {
-        "dataset": {
-            "id": 1004
-        }   
+        "id": 1004,
+        "name": "Bob",
+        "description": "Jeans",
+        "programId": 1 
     }
     return payload
--- a/api/mutations.py
+++ b/api/mutations.py
@@ -50,8 +50,6 @@ def resolve_create_dataset(obj, info, input):
     #TODO check docs- more efficient syntax to handle loading associated programs?
     #TODO check docs- is there a fancy decorator to handle automatic snakecase mapping of inputs?
 
-    print(f'{input} input')
-
     dataset_input = {
         "name": input["name"],
         "description": input["description"],
@@ -59,7 +57,6 @@ def resolve_create_dataset(obj, info, input):
         "inputter_id": '3dd10e5f-be57-435c-b721-28f3ced5dc89',
     }
 
-    print(f'{dataset_input} clean_input')
     dataset = Dataset(**dataset_input)
     session.add(dataset)
     session.commit()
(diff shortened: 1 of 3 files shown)


api/mutations.py
```diff
@@ -1,13 +1,22 @@
+import datetime
+import uuid
 from ariadne import convert_kwargs_to_snake_case, ObjectType
+from settings import settings
+from database import SessionLocal, User, Dataset, Tag, Program
+from sqlalchemy.orm import joinedload
 
+session = SessionLocal()
```
jnu: did you see an example that did something like this? typically you don't wanna create a session in a global scope like this, because then you only have one single session spanning the lifetime of the application. instead, sessions in a webapp should only last the lifetime of a single request.

In FastAPI they use this "dependency" pattern here: https://fastapi.tiangolo.com/tutorial/sql-databases/#create-a-dependency

This creates a single session for the lifetime of the request. It's a little weird to do this with Ariadne, but it looks like possibly the `info.context` object is a good place to store the session -- this is a kind of tricky bit, lmk if you wanna pair on it!

api/mutations.py
```diff
@@ -21,13 +30,53 @@ def resolve_upsert_user(obj, info, input):
...
+    return payload
+
+@convert_kwargs_to_snake_case
+@mutation.field("createDataset")
+def resolve_create_dataset(obj, info, input):
+    '''GraphQL query to create a dataset.
+        :param id: Params to be changed 
+        :returns: Newly created Dataset dictionary with eager-loaded associated Tags
+    '''
+    #TODO Flush vs Commits??
```
jnu: should be safe to just use `commit` always, since it calls `flush` for you, and not worry about the technical distinction.

(in a database transaction, you can run a lot of commands on a database to insert/delete/update records and they won't be saved by the database until you call "commit". SQLAlchemy has a similar system for modifying your data within python, which you can then _flush_ to send those operations in a transaction to the database, which will then save them when you call _commit_.)

api/mutations.py
```diff
@@ -21,13 +30,53 @@ def resolve_upsert_user(obj, info, input):
...
+    #TODO check docs- more efficient syntax to handle loading associated programs?
+    #TODO check docs- is there a fancy decorator to handle automatic snakecase mapping of inputs?
+
+    dataset_input = {
+        "name": input["name"],
+        "description": input["description"],
+        "program_id": input["programId"],
+        "inputter_id": '3dd10e5f-be57-435c-b721-28f3ced5dc89',
```
jnu: why is this one hardcoded and not coming from `input`?

api/mutations.py
```diff
@@ -21,13 +30,53 @@ def resolve_upsert_user(obj, info, input):
...
+    }
+
+    dataset = Dataset(**dataset_input)
+    session.add(dataset)
+    session.commit()
```
jnu: we should probably only call `commit` one single time, once we've added everything to the session. the reason being, if anything goes wrong partway through this update, we don't want to have a partially-completed update in the database --- we'd rather have everything _rolled back_ as a single unit.

api/mutations.py
```diff
@@ -21,13 +30,53 @@ def resolve_upsert_user(obj, info, input):
...
+            "tag_type": tag["tagType"],
+            "programs": [program],
+            "datasets": [dataset]
+        }
+        tag = Tag(**tag_input)
+        session.add(tag)
+    session.commit()
+
+    persisted_dataset = session.query(Dataset).filter(Dataset.id == dataset.id).options(joinedload("tags")).first().__dict__
+    persisted_dataset["tags"] = map(lambda tag: tag.__dict__, persisted_dataset["tags"])
```
jnu: it's almost always better/faster in python to use that _list comprehension_ syntax instead of the `map` function. so here:
```py
[tag.__dict__ for tag in persisted_dataset['tags']]
```

api/mutations.py
```diff
@@ -21,13 +30,53 @@ def resolve_upsert_user(obj, info, input):
...
+
+    dataset_input = {
+        "name": input["name"],
+        "description": input["description"],
+        "program_id": input["programId"],
+        "inputter_id": '3dd10e5f-be57-435c-b721-28f3ced5dc89',
+    }
+
+    dataset = Dataset(**dataset_input)
+    session.add(dataset)
```
jnu, line 58: as i mentioned in the other comment, the `session` really needs to be scoped to within the lifetime of a single request, and not shared between requests. (otherwise we'll end up seeing a bunch of incorrect updates and potential security flaws when we have multiple users using the app at once!!)

the simplest thing to do is just create a new `session` within each mutation and `close` it at the end.

a slightly fancier thing is to create the `session` with a dependency as in the FastAPI docs, and then put that in the `info.context` that gets shared throughout all the graphql resolvers.

jnu (changes requested):
this is awesome, nice work!!! im mainly focused on the transaction / session logic here because that can be tricky stuff that's hard to get right. if you have other things you want me to look at in particular let me know!

api/schema.graphql
```diff
@@ -102,8 +109,11 @@ type Query {
...
   # Upsert a user.
-  upsertUser(input: UpsertUserInput): UpsertUserOutput!
+  upsertUser(input: UpsertUserInput): User!
 
   # Delete a user.
-  deleteUser(id: ID!): DeleteUserOutput!
+  deleteUser(id: ID!): Int!
+
+  # Add a Dataset
+  createDataset(input: CreateDatasetInput): Dataset!
```
seerocode, line 118: Should there be a "createRecord" mutation here as well or is that a different story/PR?

jnu: there definitely should be -- i don't mind if they're all in the same pr or in different ones

api/mutations.py
nmanduj1: Interesting! Is there a way to use a session without having to pass it in as an argument at all?

api/mutations.py
nmanduj1: o! bc i was testing and have one sole user in my db- i will pull that out!

api/schema.graphql
nmanduj1: I figured I'd add them as I worked on them- "deleteUser" and "upsertUser" are here from testing the original scaffolding.

api/mutations.py
nmanduj1: fancy fancy! I will update

api/mutations.py
nmanduj1: okedokes, i added a simple session within each mutation for now-

jnu (approved):
woohoo!! this looks great!

@seerocode do you wanna review and make sure the schema changes aren't going to clash with anything you're doing in your branch?

seerocode (approved):
🚀 looks good to me, no conflict on my end

PR #33, opened 2021-04-12
Mime types and simulataneous processes

jnu:
Two small fixes:
1. start both nginx and the mock-server in the entrypoint (via docker-init.sh)
2. serve correct mime-types from nginx


seerocode (approved):
🚀

PR #36, opened 2021-04-13
[ch170] Implement "save data" buttons

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/170

The large number of file changes is sort of inevitable when you make graphql schema changes and bring them into the client when generating types for them.

### DIFF
--- a/api/schema.graphql
+++ b/api/schema.graphql
@@ -108,6 +108,28 @@ input CreateDatasetInput {
   tags: [TagInput!]!
 }
 
+input UpsertDatasetRecordInput {
+  id: ID
+  datasetId: ID!
+  publicationDate: Date
+  data: [UpsertDataInput!]!
+}
+
+input UpsertDataInput {
+  id: ID
+  category: String!
+  categoryValue: String!
+  count: Int!
+}
+
+type UpsertDatasetRecordOutput {
+  record: Record!
+}
+
+type DeleteDatasetRecordOutput {
+  id: ID!
+}
+
 type Query {
   # Retrieve a single user.
   user(id: ID!): User
@@ -126,4 +148,10 @@ type Mutation {
 
   # Add a Dataset
   createDataset(input: CreateDatasetInput): Dataset!
+
+  # Upsert a record.
+  upsertRecord(input: UpsertDatasetRecordInput): UpsertDatasetRecordOutput!
+
+  # Delete a record.
+  deleteRecord(id: ID!): DeleteDatasetRecordOutput!
 }
--- a/client/src/components/AggregateDataEntryForm.tsx
+++ b/client/src/components/AggregateDataEntryForm.tsx
@@ -13,7 +13,7 @@ const { Text } = Typography;
 interface FormProps {
   datasetId: string;
   recordId?: string;
-  formSubmitted: any;
+  onFormSubmitted: any;
 }
 
 interface Record {
@@ -80,10 +80,10 @@ const AggregateDataEntryForm = (props: FormProps): JSX.Element => {
   // TODO: query for get_dataset to EDIT a record
   const [UpsertRecord] = useMutation(UPSERT_RECORD, {
     onCompleted() {
-      props.formSubmitted(true);
+      props.onFormSubmitted(true);
     },
-    onError(error) {
-      props.formSubmitted(false); // TODO: pass error to parent
+    onError() {
+      props.onFormSubmitted(false); // TODO: pass error to parent
     },
     awaitRefetchQueries: true, // TODO: update cache instead of refetch
     refetchQueries: [
--- a/client/src/components/AggregateDataEntryForm.tsx
+++ b/client/src/components/AggregateDataEntryForm.tsx
@@ -119,7 +119,6 @@ const AggregateDataEntryForm = (props: FormProps): JSX.Element => {
 
   const handleSubmit = (e: { preventDefault: () => void }) => {
     e.preventDefault();
-    // TODO: add data/update data
     const rec = {
       input: {
         datasetId: props.datasetId,
--- a/client/src/components/AggregateDataEntryForm.tsx
+++ b/client/src/components/AggregateDataEntryForm.tsx
@@ -1,4 +1,10 @@
-import React, { ChangeEvent, useState } from "react";
+import React, {
+  ChangeEvent,
+  Dispatch,
+  FormEvent,
+  SetStateAction,
+  useState,
+} from "react";
 import { Button, Card, Col, Row, Space, Typography } from "antd";
 import { SaveOutlined, CloseSquareFilled } from "@ant-design/icons";
 import "./AggregateDataEntryForm.css";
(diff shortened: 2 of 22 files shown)


api/schema.graphql
```diff
@@ -101,6 +101,28 @@ type DeleteUserOutput {
   id: ID!
 }
 
+input UpsertDatasetRecordInput {
```
seerocode, line 111: @nmanduj1 tagging you here so we can talk about the record mutation.

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+      count: 0,
+    },
+  ],
+};
+
+const AggregateDataEntryForm = (props: FormProps): JSX.Element => {
+  // TODO: query for get_dataset to EDIT a record
+  const [UpsertRecord] = useMutation(UPSERT_RECORD, {
+    onCompleted() {
+      props.formSubmitted(true);
```
jnu: just a style note, usually you write these callbacks with names like `onFormSubmitted`

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+    },
+    onError(error) {
+      props.formSubmitted(false); // TODO: pass error to parent
```
jnu: a typical syntax you'll find for event callbacks is `onFormSubmitted(error: Error | undefined)`. If there was an error, you'll pass it in, otherwise you'll assume things succeeded.

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+    tempValues.data[index] = {
+      ...tempValues.data[index],
+      count: +event.target.value,
+    };
+    setValues(tempValues);
+  };
+
+  const handleSubmit = (e: { preventDefault: () => void }) => {
+    e.preventDefault();
+    // TODO: add data/update data
```
jnu: what is this todo referring to?

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+      <div className="data-entry_publication-date-field">
+        <label htmlFor="publicationDate">
+          <Text>Released on: </Text>
+        </label>
+        <input
+          type="date"
+          id="publicationDate"
+          name="publicationDate"
+          required
+          onChange={(e) => setPublicationDate(e.target.value)}
```
jnu, line 157: just checking, does this actually set the value as a `string` or is it a `Date` object?

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+        />
+      </div>
+      <Row gutter={[16, 16]} className="data-entry">
+        <Col span={8}>
+          <h3 className="data-entry_category-descr-header">About gender</h3>
```
jnu: This header should come from the translations file

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+              <Button
+                type="primary"
+                htmlType="submit"
+                icon={<SaveOutlined />}
+                style={{ whiteSpace: "normal", height: "auto" }}
+              >
+                Save Record
+              </Button>
+              <Button
+                onClick={() => history.push("/")}
```
jnu, line 205: hmm -- why is this pushing the root `/` path? shouldn't it go to the dataset overview page? also, i don't think we really want to push a new history state here, but rather `pop` the current one (right? just thinking about how the back button should work when we click this).

client/src/components/AggregateDataEntryForm.tsx
```diff
@@ -1,65 +1,207 @@
...
+                icon={<CloseSquareFilled />}
+                style={{ whiteSpace: "normal", height: "auto" }}
+              >
+                Cancel and Return To Dashboard
```
jnu: This (and other button text) should come from the translations file

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -110,7 +109,9 @@ const DatasetRecordsTable = ({
       size="small"
       scroll={{ x: 1000 }}
       sticky
-      title={() => "Summary"}
+      title={() => "Records"}
+      pagination={{ pageSize: 6 }}
```
jnu: just curious, why 6?

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -110,7 +109,9 @@ const DatasetRecordsTable = ({
       size="small"
       scroll={{ x: 1000 }}
       sticky
-      title={() => "Summary"}
+      title={() => "Records"}
```
jnu: why is the title in a function? also can these titles come from the translations file? :)

jnu (approved):
this looks great! i made a few suggestions to clean up but really nothing major. This is really nice -- great work ro!!!

client/src/components/AggregateDataEntryForm.tsx
seerocode: Changed.

client/src/components/AggregateDataEntryForm.tsx
seerocode: There wasn't a ticket for editing data and this was a TODO to accomplish that in this handler. Edited to remove TODO.

client/src/components/AggregateDataEntryForm.tsx
seerocode: Sets it as a string.

client/src/components/AggregateDataEntryForm.tsx
seerocode: Done.

client/src/components/AggregateDataEntryForm.tsx
seerocode: I checked the [docs](client/src/components/AggregateDataEntryForm.tsx) and pop doesn't exist (replace does). I initially had a router Link wrapping each button but learned that was [not valid HTML](https://validator.w3.org/nu/?showsource=yes&doc=http%3A%2F%2Fdropbox.beausmith.com%2Fstatic-html%2Finvalid-button-anchor-markup.html).

Pushing a link onto the stack looked like the next option. What would be a better way?

client/src/components/AggregateDataEntryForm.tsx
seerocode: Done.

client/src/components/DatasetRecordsTable.tsx
seerocode: Moved to translations file. The `title` property of an ant design table expects a function as a value.

client/src/components/DatasetRecordsTable.tsx
seerocode: The table spilled over the parent container when it exceeded six rows. Seems to be an issue with the footer and table interacting (tried CSS changes) but since we're going to be changing this table in the future, I felt this was a good compromise for now.

client/src/components/AggregateDataEntryForm.tsx
seerocode: @jnu update component to pass error up to parent here: https://github.com/stanford-policylab/bbc-50-50/pull/36/commits/6bec31e8c4c3bc06571b072ccc4590c0555070d8

Is this closer to your recommendation?

PR #38, opened 2021-04-20
Ch182/query users

nmanduj1:


api/queries.py
```diff
@@ -16,27 +15,19 @@ def resolve_user(obj, info, id):
...
-    return payload
+    return retrieved_user
 
 @query.field("dataset")
 @convert_kwargs_to_snake_case
 def resolve_dataset(obj, info, id):
     '''GraphQL query to find a dataset based on dataset ID.
         :param id: Id for the dataset to be fetched
-        :returns: Dataset dictionary with eager-loaded associated Tags and Records
+        :returns: Dataset dictionary
```
jnu, line 28: worth noting this is _not_ a dictionary anymore, but a `Dataset` object. same with the docs for the `User` resolver above!

jnu (approved):
woohoo!! looks great

PR #39, opened 2021-04-21
[ch174] Delete record when user clicks "delete" button on dataset details page

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/174

**Changes:**
- Added mutation to delete a record
- Added error and loading after a record is deleted
- Added confirmation notice on record deletion
- Added test for user action on delete which would trigger the mutation and error states
- Added a custom mocked provider using Apollo Client (because Apollo's Mocked provider does not work to test refetch a query after a mutation)


seerocode:
I branched off of `ch170/save-data` and will rebase when that's merged which will clean ups most of these commits and file changes.

seerocode:
Working on test..stand by

client/src/components/DatasetRecordsTable.tsx
```diff
@@ -1,37 +1,69 @@
...
 }: DatasetRecordsTableProps): JSX.Element => {
-  const { data, loading, error } = useQuery<GetDataset, GetDatasetVariables>(
-    GET_DATASET,
-    {
-      variables: { id: datasetId },
-    }
-  );
+  const { t } = useTranslation();
+
+  const tableData = records?.map((record) => {
```
seerocode: Should this data transformation from the query result be moved up to the parent component instead of being passed down as a prop for the child component to handle? Or is there some more logical place to put it?

client/src/__mocks__/AutoMockProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  // 5) Define ApolloClient (client variable used below)
+  const client = new ApolloClient({
+    link: splitLink,
+    cache: new InMemoryCache(),
+  });
+
+  return <ApolloProvider client={client}>{children}</ApolloProvider>;
+};
+
+export { AutoMockedProvider };
```
jnu: I'm curious if you found some docs that suggested to set things up this way? it's a clever idea, my gut reaction is generally that you shouldn't create objects like this during a component's render cycle, since they can get destroyed and recreated unpredictably

my impulse instead would be to move the `apolloClient` creation outside of the render function and just use the normal `ApolloProvider`, like this:
```ts
const autoMockedClient = ({customResolvers, errors}) => {
  // create the client here
};

test('it does stuff', () => {
  const clietn = autoMockedClient(...);
  render(`<ApolloProvider client={client}>...</ApolloProvider>`)
});
```

this is true for the loading provider as well

jnu (approved):
looks great to me!

client/src/components/DatasetRecordsTable.tsx
jnu: i could go either way. i think it makes sense to encapsulate transformations for the purpose of presentation as close to where that presentation happens as possible, so in the child. some reasons you _wouldn't_ want to do that include: 1) performance, if it's an expensive transformation, and 2) DRYness, if other components would also benefit from sharing that transformation. 

if you move the transformation to the parent, your parent/child become more tightly coupled and interdependent on each other, which is generally _not_ good design.

client/src/__mocks__/AutoMockProvider.tsx
seerocode: Yeah, a few with the same approach:
- https://www.telerik.com/blogs/mocking-and-testing-graphql-in-react
- https://www.youtube.com/watch?v=VjmAzW2lrR8
- https://www.freecodecamp.org/news/a-new-approach-to-mocking-graphql-data-1ef49de3d491/

I found them after not finding a good solution for mocking a change to the UI after deleting a record. The mutation triggers a refetch of the dataset query on delete and Apollo's MockedProvider doesn't handle this currently although there have been issues requesting that feature. There's a property called "newData" that's supposed to trigger a refetch but so far online, the only way to get this to work is with a boolean hack.

A custom mock made more sense to me while still being able to use Apollo's mocked provider if needed. In general, Apollo's version worked super well as long as you didn't need to refetch any data but it was also annoying to have to mock a return for EVERY single query and mutation whereas this custom mock, creates mock for you if one hasn't been provided for a "string" field, for example. Much like our current mock server.

seerocode: Also, I agree with taking the client out of the render! I'll switch that.

client/src/components/DatasetRecordsTable.tsx
seerocode: Thank you! I was worried this didn't make sense the way I wrote it.

PR #42, opened 2021-04-22
[ch194] Dummy data for database

jnu:
Makes some mock data based on https://github.com/stanford-policylab/bbc-50-50/pull/25/commits/b6e5022e7e7e11b2b8588663e48ce24b680aecb7

Added some defaults for the columns as well so they're easier to work with.

Create the mock data with:
```
python database.py --no-tables --dummy-data
```
(you can omit the `--no-tables` flag to simultaneously create the tables)

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -1,6 +1,7 @@
 import databases
 import uuid
 
+import click
 from sqlalchemy import create_engine, Table, Boolean, Column, Integer, Float, String, DateTime, ForeignKey
 from sqlalchemy.orm import relationship, sessionmaker
 from sqlalchemy.ext.declarative import declarative_base
@@ -56,7 +57,7 @@ class Organization(Base):
 
     id = Column(GUID, primary_key=True)
     name = Column(String(255), nullable=False)
-    teams = relationship('Team')
+    teams = relationship('Team', back_populates='organization')
 
     created = Column(TIMESTAMP,
                      server_default=func.now(), nullable=False)
@@ -193,6 +194,7 @@ class Record(Base):
     __tablename__ = 'record'
 
     id = Column(GUID, primary_key=True)
+    dataset = relationship('Dataset', back_populates='records')
     dataset_id = Column(GUID, ForeignKey(
         'dataset.id'), nullable=False, index=True)
     publication_date = Column(DateTime)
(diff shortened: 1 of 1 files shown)


api/database.py
```diff
@@ -54,7 +63,7 @@
 class Organization(Base):
     __tablename__ = 'organization'
 
-    id = Column(GUID, primary_key=True)
+    id = Column(GUID, primary_key=True, default=uuid.uuid4)
```
seerocode, line 66: Does it make a difference if it's "server" vs "server_default"?

api/database.py
```diff
@@ -206,12 +216,58 @@ class Record(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
-if __name__ == '__main__':
+
+@click.command()
```
seerocode, line 220: I had to look up this click package and it is so cool!! How did you find out about it?

api/database.py
```diff
@@ -206,12 +216,58 @@ class Record(Base):
...
-        name="admin", description="User is an admin and has administrative privileges"))
-    session.commit()
+        session.add(Role(
+            id="be5f8cac-ac65-4f75-8052-8d1b5d40dffe",
+            name="admin",
+            description="User is an admin and has administrative privileges"))
+        session.commit()
+
+    if dummy_data:
+        print("👩🏽‍💻 Adding dummy data ...")
```
seerocode, line 239: This is organized so well! Thanks for putting this together.

seerocode (approved):

api/database.py
jnu: there's a difference! `default` specifies a function that will be executed in python, `server_default` asks the database itself to run a function. in practice, either one is fine for our application, but generally the `server_default` is a bit better because it'll behave the same if you're interacting with the database in another tool (outside of python). however, uuids are a little weird because lots of databases (including postgres) don't natively include uuid-generation functions themselves, which is why i went with `default` here.

api/database.py
jnu: i've used in before in a few projects over the years! i forget where i first came across it. it's usually worth googling "CLI framework" in any language you're using, because these things are tedious to write by hand (and python's native `argparse` library is also fairly annoying to use).

PR #46, opened 2021-04-23
Ch199/update dataset

nmanduj1:


api/database.py
```diff
@@ -151,6 +151,10 @@ class Tag(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
+@event.listens_for(Tag, 'before_insert')
+def capitalize_tag_name(mapper, connect, target):
+    # target is an instance of Table
+    target.name = target.name.capitalize().strip()
```
seerocode, line 157: The tags are capitalized on the UI for you! I had used the example table here: https://ant.design/components/table/

api/mutations.py
```diff
@@ -89,4 +89,43 @@ def resolve_delete_dataset(obj, info, id):
...
+def resolve_update_dataset(obj, info, input):
+    '''GraphQL mutation to update a Dataset.
+        :param input: Params to be changed
+        :returns: Updated Dataset
+    '''
+    session = info.context['dbsession']
+    dataset = session.query(Dataset).filter(Dataset.id == input["id"]).first()
+
+    # removing tags from input, and returning the contents of tags OR an empty array if tags was not found.
+    tags = input.pop('tags', [])
```
seerocode, line 105: Oh, cool! I hadn't accessed an array like this before. Is this the same as assigning `input["tags"]` to `tags` or does it do something different?

api/mutations.py
```diff
@@ -89,4 +89,43 @@ def resolve_delete_dataset(obj, info, id):
...
+
+    # iterating over tags to check if the included tag exists or should be created.
+    for tag in tags:
```
seerocode, line 108: You wrote this out in such a clear way! I was thinking maybe it doesn't matter if the tag exists but then I remembered that replacing already existing tags could render a new ID and cause stuff to break.

api/database.py
nmanduj1: Nice! I like your ui decision! I'd like a check on the backend as well just to make sure I'm not accidentally duplicating / missing tags because of their case or if we ever expose our api. Also, .capitalize() only capitalizes the first letter (you probably knew this- i was surprised bc i was expecting it to capitalize everything !! ).

api/mutations.py
nmanduj1: Yea! I was excited when i found it too- .pop removes the item at a specific key and returns it, the second argument, [], specifies what to return in case the key is not found. In this way, I can always expect an array and it wont break my logic down the line. pretty handy!

api/database.py
seerocode: Oh yes makes sense! I didn't even know Python had capitalize. I learned it with JavaScript having to do the same thing.

seerocode (approved):
🚢

api/database.py
jnu: hey sorry i'm late here, fyi this is probably better to do with validators in sqlalchemy rather than events -- just a little bit tidier. also just noting that for internationalization, capitalization gets a little weird

api/mutations.py
```diff
@@ -89,4 +89,43 @@ def resolve_delete_dataset(obj, info, id):
...
+        existing_tag = session.query(Tag).filter(Tag.name == tag["name"].capitalize().strip()).one_or_none()
```
jnu, line 109: another tiny thing, but since you do this `.capitalize().strip()` standardization logic in multiple places it should be encapsulated in a function (since if you change it in one place we wanna make sure it's changed in both places!)

api/database.py
nmanduj1: I am not looking to validate the incoming format. I am using this as a way to normalize the tag names before persisting to the db. i don't want to throw an error just because the tag name was not formatted "properly." Thoughts? Would a validator still be a good tool for this use case?

jnu: yep! you don't have to raise an error in the validator if you don't want to, you can just return the properly formatted value silently. see more detail here: https://docs.sqlalchemy.org/en/14/orm/mapped_attributes.html#simple-validators

there are three reasons validators are usually used for this instead of events:
1) to keep the property formatting logic consolidated in the same class (`Tag` here),
2) to ensure the object's property value (`tag.name`) has the correctly formatted value as soon as you create the object, instead of waiting until `session.flush` is called (in case you reference it within the session before you get to the flush step), and
3) to guard against potentially unsafe operations in the mapper event handler (see the warnings in the [docs](https://docs.sqlalchemy.org/en/14/orm/session_events.html#mapper-level-events)).

the end result in the database is the same for either approach, but it's good to know about the differences (i could see number 2 coming up at some point, for example). just suggesting it as an alternative!

nmanduj1: Okedokes, thanks!

PR #47, opened 2021-04-26
[ch168] Implement "edit record" button - clean up and setup 

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/168

Cleaned up mutations and components to set up for adding a new mutation to edit a record. Following PRs will be set up to be connected to the API instead of mock-server.


seerocode:
Fixing test!

api/schema.graphql
```diff
@@ -108,6 +108,7 @@ input CreateDatasetInput {
   tags: [TagInput!]!
 }
 
+<<<<<<< HEAD
```
jnu: looks like you've got an unfinished rebase here?

api/schema.graphql
```diff
@@ -118,21 +119,24 @@ input UpdateDatasetInput {
...
   id: ID
   category: String!
   categoryValue: String!
   count: Int!
 }
 
-type UpsertDatasetRecordOutput {
-  record: Record!
+type InsertDatasetRecordOutput {
+  id: ID!
```
jnu, line 135: you don't necessarily have to make a new `type` for the output of a mutation. typically for `insert` mutations it makes sense just to return the object you inserted (in this case a `Dataset`), so that your client has maximum flexibility about what they read as a response to the mutation

jnu (approved):
looks great, nice work ro!! not sure what's going on with the unresolved conflict tho - you'll have to resolve that before merging!

api/schema.graphql
seerocode: oh, yup! 🤦🏽

api/schema.graphql
seerocode: Nancy had given me the same advice! I was going to remove it in the next PR to avoid the nesting.

jnu: ok awesome sounds good!!

nmanduj1:
🚢

PR #48, opened 2021-04-28
[ch168] Implement "edit record" button

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/168

Please ignore the changes in files in the _generated_ directory as these are auto generated types by apollo.

### 🔑 changes (note that adding and deleting were closely tied together and implementing edit meant updating the add feature):
- Can edit now! 
image
- Buttons change if you're editing or adding
image
image
- Alerts on error when submitting the form
image
- Different success message on form submission for edit and add
image
image
- Form resets and toast success appears when you save and add another record (as suggested in a Slack thread)
image
- Added a wrapper around ant design message to handle toast notifications ([commit](https://github.com/stanford-policylab/bbc-50-50/pull/48/commits/9d955831017e6871cb703faeb01ee9f72850c546))
- See further comments on changes in conversation below


api/schema.graphql
```diff
@@ -117,26 +117,32 @@ input UpdateDatasetInput {
...
+input UpdateRecordInput {
   id: ID!
+  datasetId: ID!
+  publicationDate: Date!
+  entries: [UpdateEntryInput!]!
 }
 
-type DeleteDatasetRecordOutput {
+input UpdateEntryInput {
   id: ID!
```
seerocode: @nmanduj1 this differs in your PR ([line here](https://github.com/stanford-policylab/bbc-50-50/pull/49/files#diff-00a344d86c0b9c44781ec75c4caaec1017852edaed72802e6a37e184723f9a1fR133)) only in requiring the ID when updating a record so we know what ID record to update. Does that make sense?

api/schema.graphql
```diff
@@ -162,12 +171,16 @@ type Mutation {
...
+  # Create a record
+  createRecord(input: CreateRecordInput!): Record!
+
+  # Update a Record
+  updateRecord(input: UpdateRecordInput!): Record!
 
-  # Delete a record.
-  deleteRecord(id: ID!): DeleteDatasetRecordOutput!
+  # Delete a record
+  deleteRecord(id: ID!): ID!
```
seerocode, line 171: aside: I wonder if this should return the dataset for the ID instead so we don't have to refetch the dataset query to update the table every time a record is deleted?

see: https://www.apollographql.com/blog/when-to-use-refetch-queries-in-apollo-client/

api/schema.graphql
```diff
@@ -82,10 +82,10 @@ type Record {
   id: ID!
   dataset: Dataset!
   publicationDate: Date!
-  data: [Data!]!
+  entries: [Entry!]!
 }
 
-type Data {
+type Entry {
```
seerocode: Made this update to coincide with PR #49

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
-        subTitle={`Sorry, something went wrong. ${formState?.errors?.message}`}
-        extra={
-          <Button type="primary" onClick={() => location.push("/")}>
-            {t("backToDashboard")}
-          </Button>
-        }
-      />
-    );
-  }
+  const isEditMode = recordId ? true : false;
```
seerocode, line 30: This will be true when a user edits a record as it will use the "edit" route that has a :datasetId and :recordId param.

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+
+  const programAndDataset = `${datasetData?.dataset.program.name} | ${datasetData?.dataset.name}`;
+
+  /* Retrieve a record to populate the data entry form if editing */
+  const {
+    data: existingRecord,
+    called: existingRecordCalled,
+    loading: existingRecordLoading,
+    error: existingRecordError,
+  } = useQuery<GetRecord, GetRecordVariables>(GET_RECORD, {
```
seerocode, line 49: We could use `lazyQuery` instead but implementation-wise, this works the same and logically made more sense to me.

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+  if (datasetLoading || (existingRecordCalled && existingRecordLoading))
+    return <h1>Loading data...</h1>;
+  if (datasetError || (existingRecordCalled && existingRecordError))
+    return <h1>Error loading data. Please refresh and try again.</h1>;
+
   return (
     <>
       <Title style={{ marginBottom: "10px" }} level={2}>
-        Add record for{" "}
+        {pageTitle} record for{" "}
```
seerocode: This will change to `edit` or `add` depending on edit mode.

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
         <Link
           to={{
             pathname: `/dataset/${datasetId}/details`,
           }}
         >
-          {"BBC News - Breakfast Hour"}
+          {` ${programAndDataset}`}
         </Link>
       </Title>
-      <DataEntryPersonTypesInput />
```
seerocode, line 83: Removed this from the MVP for now while we make updates to schema to accommodate a person type.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -5,96 +5,99 @@ import React, {
...
-      category: "gender",
-      categoryValue: "gender non-conforming",
-      count: 0,
-    },
+interface Category {
+  categoryType: string;
+  categoryAttributes: string[];
+}
+
+const categories: Category = {
```
seerocode, line 42: This is my first pass at what a Category on a dataset would look like. Not important now but wanted to point this out as I removed the default form.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -103,48 +106,144 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
...
+
+    try {
+      const data = await create();
+      if (data) props.onFormSubmitted(true);
+    } catch (err) {
+      setError(err);
+    }
+  };
+
+  const handleSubmitReload = async () => {
```
seerocode, line 185: This function is only triggered when a user wants to submit a new record and stay on the same page.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -103,48 +106,144 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
...
+
+  // TODO: prevent user from editing date
   return (
-    <form onSubmit={handleSubmit}>
+    <form onSubmit={handleSubmit} id="data-entry-form" aria-label="Data Entry">
+      {createError && error && (
+        <Alert
+          message="Oh, no! Something went wrong"
+          description={
+            error.message.includes("duplicate key value")
```
seerocode, line 229: There should be a global error handler for this but I left it here for now. This error returns if a user is attempting to edit a new record for a date that already exists for the dataset. It changes the wordy psycopg error to something more readable.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -153,55 +252,101 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
...
                   className="data-entry-form_label"
                 >
                   <input
                     name={item.categoryValue}
                     required
-                    type="number"
-                    min="0"
+                    aria-labelledby={item.categoryValue}
+                    aria-required="true"
+                    type="text"
```
seerocode, line 292: Changed this from number to text input for accessibility reasons. See: https://www.coywolf.news/webmaster/accessible-number-input/

api/schema.graphql
nmanduj1: I think I see what you're saying- From our conversations, we decided updating a record could handle both creating and updating an entry. To do that, the ID should be optional- the api can figure out if the id is present for each entry and knows to create or update.

seerocode: If it's just required for the update entry input and not the add entry input on the schema side, that should still work right?

api/schema.graphql
nmanduj1: This is interesting! It would be so helpful to learn how apollo handles caching. I would be inclined to update the cache by removing the uuid and having that update to state trigger a reload of the component. But it looks like apollo requires a refetch (to the api??) ?  I'd love to talk this through with you!

client/.eslintrc.js
```diff
@@ -17,9 +17,11 @@ module.exports = {
     "plugin:@typescript-eslint/recommended", // Uses the recommended rules from @typescript-eslint/eslint-plugin
     "plugin:jsx-a11y/recommended"
   ],
-  plugins: ["jsx-a11y"],
+  plugins: ["jsx-a11y", "react-hooks"],
   rules: {
     // Place to specify ESLint rules. Can be used to overwrite rules specified from the extended configs
     // e.g. "@typescript-eslint/explicit-function-return-type": "off",
+    "react-hooks/rules-of-hooks": "error",
```
ameliagoodman, line 24: I'm just curious what this does?

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+  if (formSuccess)
     return (
       <Result
         status="success"
-        title="Success!" // TODO: Render dynamically
-        subTitle="Data saved for BBC News - Breakfast Hour"
+        title="Success!"
+        subTitle={`Data has been saved succesfully for ${programAndDataset}`}
         extra={[
+          !isEditMode && (
```
ameliagoodman, line 63: im a lil confused about what's going on here with the `&&`

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+            <Button
```
ameliagoodman, line 64: for my own sanity, where are you importing `Button`

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
-            onClick={() => location.push(`/dataset/${datasetId}/entry/reload`)}
-          >
-            {t("addMoreData")}
-          </Button>,
-          <Button
+            type={isEditMode ? "primary" : undefined}
             key="goToDataset"
             icon={<DashboardTwoTone />}
-            onClick={() => location.push(`/dataset/${datasetId}/details`)}
+            onClick={() => history.push(`/dataset/${datasetId}/details`)}
```
ameliagoodman, line 77: is history.push just like settting the router?

client/src/components/DataEntry/DataEntry.tsx
ameliagoodman: why have `{" "}`

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -26,9 +27,10 @@ const DatasetDetailsRecordsTable = ({
   isLoading,
 }: DatasetRecordsTableProps): JSX.Element => {
   const { t } = useTranslation();
+  const history = useHistory();
 
   const tableData = records?.map((record) => {
-    return record.data.reduce(
+    return record.entries.reduce(
```
ameliagoodman, line 33: the nested map / reduce  makes my brain hurt,.. i could use a comment describing like how input and output compare

api/schema.graphql
jnu: i believe you should be able to use the `update` callback to remove the dataset from the cache on delete - have you looked into that yet?

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
ameliagoodman: if it's not too much effort

client/src/components/DataEntry/DataEntryMessageHandler.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if (isLoading)
+    return messageLoading({
+      content: `Saving ${messageText} record...`,
+      key: messageKey,
+    });
+
+  if (isSuccess)
+    messageSuccess({
```
ameliagoodman: do you need the return here (same as line 20?)

client/src/components/DataEntry/DataEntry.tsx
jnu: just so u know you'll often see this written as `const isEditMode = !!recordId`. i don't mind which you use, but you might encounter that in frontend code!

client/src/index.tsx
```diff
@@ -30,8 +30,22 @@ const httpLink = new HttpLink({
       : "/graphql/",
 });
 
+const cache = new InMemoryCache({
+  typePolicies: {
+    Dataset: {
+      fields: {
+        records: {
+          merge(existing = [], incoming: any[]) {
```
ameliagoodman, line 38: I don't know whats goin on here; im not understanding the syntax. i could use a comment. are you defining merge to return only incoming?

ameliagoodman (commented):
wow what a joy it is to review this pr. you are so good !!!! mostly these are clarifying q's :)

client/src/components/DataEntry/DataEntry.tsx
jnu: yep, i agree, lazy-loading is usually an optimization that you only add once you start seeing issues!

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+  const programAndDataset = `${datasetData?.dataset.program.name} | ${datasetData?.dataset.name}`;
+
+  /* Retrieve a record to populate the data entry form if editing */
+  const {
+    data: existingRecord,
+    called: existingRecordCalled,
+    loading: existingRecordLoading,
+    error: existingRecordError,
+  } = useQuery<GetRecord, GetRecordVariables>(GET_RECORD, {
+    variables: { id: recordId! },
```
jnu, line 50: nice use of `!` :D

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+            onClick={() => history.push(`/dataset/${datasetId}/details`)}
           >
             {t("goToDataset")}
           </Button>,
         ]}
       />
     );
 
+  if (datasetLoading || (existingRecordCalled && existingRecordLoading))
+    return <h1>Loading data...</h1>;
```
jnu, line 86: does antd have a nicer spinner you can use? something on here: https://ant.design/components/spin/

it's also a common pattern to refactor your loading / error states into a container, like:

```tsx
<ShmancyContainer
  error={datasetError || (existingRecordCalled && existingRecordError)}
  loading={datasetLoading || (existingRecordCalled && existingRecordLoading)}>
   {/* all the children */}
</ShmancyContainer>
```

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+  if (datasetError || (existingRecordCalled && existingRecordError))
```
jnu, line 87: nice use of `|| ... &&` -- this is a tricky construction!!

client/src/components/DataEntry/DataEntry.tsx
jnu: nice! is there a reason we're not using the i18n file for this tho?

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
-        subTitle="Data saved for BBC News - Breakfast Hour"
+        title="Success!"
+        subTitle={`Data has been saved succesfully for ${programAndDataset}`}
         extra={[
+          !isEditMode && (
+            <Button
+              type="primary"
+              key="addMoreData"
+              icon={<PlusCircleTwoTone />}
+              onClick={() => history.push(`/dataset/${datasetId}/entry/reload`)}
```
jnu, line 68: i think im missing something, what is the /reload route?

client/.eslintrc.js
seerocode: It lints any react hooks you use so that if you try adding a hook outside of a React function, it will yell at you to get it together.

client/src/components/DataEntry/DataEntry.tsx
nmanduj1: also, why history vs location?

client/.eslintrc.js
nmanduj1: same ^

client/src/components/DataEntry/DataEntry.tsx
jnu: (does it force the apollo-client to hard-refresh the page? if that's the intent, i think typically the pattern is to use the `update` key on the `useMutation` options to update the cached data!) see here: https://www.apollographql.com/docs/react/data/mutations/#making-all-other-cache-updates

client/src/components/DataEntry/DataEntry.tsx
seerocode: I learned this from here!!: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

If the user is not in edit mode, then it will render whatever is after the `&&` operator. In this case it's saying that it will render an extra button to "add more data" to the success screen.

client/src/components/DataEntry/DataEntryMessageHandler.tsx
```diff
@@ -0,0 +1,30 @@
...
+  isSuccess = false,
+  isLoading = false,
+  messageKey = "",
+  errorMessage = "",
+}) => {
+  const messageText = isEditMode ? `updated` : `new`;
+
+  if (errorMessage) {
+    return messageError({
+      content: `Oh, no! Something went wrong. ${errorMessage}.`,
```
jnu, line 14: thanks for adding these messages!!! can they come from i18n tho?

client/src/components/DataEntry/DataEntry.tsx
seerocode: `[Button]`(https://ant.design/components/button/) is an ant design component that allows you to add icons and make buttons a little fancy without extra work. They are also accessible which was great when testing for it.

client/src/components/DataEntry/DataEntry.tsx
```diff
@@ -15,76 +21,89 @@ interface RouteParams {
...
+  const [formSuccess, setFormSuccess] = useState<boolean>(false);
+
+  if (formSuccess)
     return (
       <Result
         status="success"
-        title="Success!" // TODO: Render dynamically
-        subTitle="Data saved for BBC News - Breakfast Hour"
+        title="Success!"
+        subTitle={`Data has been saved succesfully for ${programAndDataset}`}
```
jnu, line 61: can this come from i18n? :)

client/src/components/DataEntry/DataEntry.tsx
ameliagoodman: omg 🙀🙀🙀🙀 the amount of times i have { bool ? <some component> : <></> } !!!!!! this is huge!!!!

jnu (approved):
this looks great, nice job ro!! i made a few small style suggestions, nothing major. i had a couple of questions about the mutations and caching as well -- overall this looks amazing, awesome work!!!

client/src/components/DataEntry/DataEntry.tsx
seerocode: Yeah! It adds this route to the history stack and routes to the component we configured for it. The other option would be to wrap a button in a router `<Link />` which I had started with but learned that was an anti-pattern. Pushing history this way allows you still be able to go back and forth without winding up at weird locations.

client/src/components/DataEntry/DataEntry.tsx
ameliagoodman: but like, you don’t need to import it in the file?

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: Definitely! If there's a more concise and readable way to do this, I would prefer that also. I struggled trying to figure out what would make the most sense.

client/src/components/DataEntry/DataEntry.tsx
seerocode: For a space but now that I see it again, I can use a string literal instead. I'll change that.

ameliagoodman: I noticed u also include a space before the programanddataset (line 99) so u might be good to leave it out since u cover it there?

client/src/components/DataEntry/DataEntryMessageHandler.tsx
seerocode: Yup! I'll fix.

client/src/components/DataEntry/DataEntry.tsx
ameliagoodman: or u can just do {programAndDataset} on line 99

client/src/index.tsx
seerocode: This was to resolve an Apollo cache warning. When a record was deleted from the records table for a dataset, Apollo cache didn't know how to reconcile the change but would still return the correct data and log a warning that a merge function needs to be created to appropriately update the cache. 

This highlights the original warning: https://github.com/apollographql/apollo-client/issues/6451

I wonder now if cache.modify() might be more appropriate than setting this type policy? The documentation for cache modify is here: https://www.apollographql.com/docs/react/caching/cache-interaction/

client/src/components/DataEntry/DataEntry.tsx
seerocode: Yes! There is one. Ahh, this is a better option. Sort of like an error boundary? I can do that real quick.

seerocode:
@jnu I can't comment on your cache [comment](https://github.com/stanford-policylab/bbc-50-50/pull/48#discussion_r627653083) for some reason. I raised a question with @ameliagoodman here too based on her question with the cache type policy. I wonder if modifying or writing to the cache would be better? reference: https://www.apollographql.com/docs/react/caching/cache-interaction/

client/src/components/DataEntry/DataEntry.tsx
seerocode: For the cache updates: I thought that's what the type policy in the index file was doing but I can see why focusing the update on the mutation itself would be better. 

The `/reload` route is to reload the data entry form page after success.

api/schema.graphql
seerocode: That's what's suggested @jnu. Although I think the refetch command on a mutation combined with the type policy on the in memory cache does do the update?

client/src/components/DataEntry/DataEntry.tsx
seerocode: Maybe it's not showing that change @ameliagoodman ? L99 should be : ``` {` ${programAndDataset}`} ```

client/src/components/DataEntry/DataEntry.tsx
seerocode: It's so cool when you only care about one side of the boolean!

client/src/components/DataEntry/DataEntry.tsx
seerocode: Oh yeah, it should be at the top as an import: https://github.com/stanford-policylab/bbc-50-50/pull/48/files/a09f7cefcbefda0f8d0b6218ed14be8afd7c56de#diff-fa3436a0210044b064211cdfb393ec20581acd43481eea44983ea900e83053b2R1

I think it was hidden.

client/src/components/DataEntry/DataEntry.tsx
seerocode: About the only thing discrete math was useful for was drilling truth tables. 😆

nmanduj1 (approved):
🚢

client/src/components/DataEntry/DataEntry.tsx
seerocode: I'm going to come back to this in another PR so I can merge this in for now.

client/src/components/DataEntry/DataEntry.tsx
seerocode: Will also refactor these strings to come from i18n in a separate PR!

PR #49, opened 2021-04-28
Crud Records

nmanduj1:
Some general cleanup along with CRUD for Records. Also updated createDataset resolver to use our swanky new db session.

Also, i have to look into Ariadne's case mapping- it isn't working as expected. I implemented a quick hack using regex to get this out for the MVP / our meeting next week. This should be updated to use Ariadne's case mapping as an elegant solution.


api/mutations.py
```diff
@@ -129,3 +102,82 @@ def resolve_update_dataset(obj, info, input):
...
+            "category": data["category"],
+            "category_value": data["categoryValue"],
+            "count": data["count"]
+        }
+        new_data = Entry(**data_input)
+        record.entry.append(new_data)
+
+    for param in input:
+        # snake_case mapping 
+        under_scored = re.sub(r'(?<!^)(?=[A-Z])', '_', param).lower()
```
seerocode: Can you give me an example of what this is doing or solving for? I'm not sure I understand.

api/database.py
```diff
@@ -210,7 +210,7 @@ class Record(Base):
     dataset_id = Column(GUID, ForeignKey(
         'dataset.id'), nullable=False, index=True)
     publication_date = Column(DateTime, index=True, unique=True)
-    data = relationship('RecordData')
```
seerocode, line 213: Thanks for bringing this up! Def makes it more clear.

api/schema.graphql
```diff
@@ -141,33 +147,34 @@ type DeleteDatasetRecordOutput {
...
-  deleteDataset(id: ID!): String!
+  deleteDataset(id: ID!): ID!
+
   # Update a Dataset
   updateDataset(input: UpdateDatasetInput): Dataset!
 
-  # Upsert a record.
-  upsertRecord(input: UpsertDatasetRecordInput): UpsertDatasetRecordOutput!
+  # Create a Record
+  createRecord(input: CreateRecordInput): Record!
```
seerocode, line 173: I had "inputRecord" but I like this better.

api/schema.graphql
```diff
@@ -117,14 +117,20 @@ input UpdateDatasetInput {
...
+  datasetId: ID
   publicationDate: Date
-  data: [UpsertDataInput!]!
+  data: [DataInput!]
+}
+
+input CreateRecordInput {
+  datasetId: ID!
+  publicationDate: Date!
+  data: [DataInput!]
```
seerocode: Should this be EntryInput to align with the DB table change too?

api/mutations.py
nmanduj1: Graphql is handing the resolver camel cased arguments but the resolver should be working with underscored arguments. (( In this case specifically, publicationDate vs publication_date. )) 

Ariadne should handle the conversion via their decorator but it is not working as expected. I have to do some digging but didn’t want to hold us up. I have a feeling it might have something to do with the input but am not certain. 

This solution, while not elegant- works and allows us to insert Records into the db but i will have to get that decorator working!

api/schema.graphql
nmanduj1: Funny that you picked up on this bc I was debating ! i kinda like using data here but you're right, for the sake of continuity and clarity- I will update  🙏

api/mutations.py
jnu: maybe you can try flipping the order of the decorators? in other words, on line 139 you have:

```py
@convert_kwargs_to_snake_case
@mutation.field("updateRecord")
def resolve_update_record(obj, info, input):
```

and it should be:

```py
@mutation.field("updateRecord")
@convert_kwargs_to_snake_case
def resolve_update_record(obj, info, input):
```

the `@` syntax is just shorthand for composing functions. the first thing you have is equivalent to writing:
```py
convert_kwargs_to_snake_case(mutation.field("updateRecord")(resolve_update_record))
```
which would mean the convert_kwargs function is getting applied to the wrapped resolver, not the resolver itself!

lmk if that helps!

api/mutations.py
```diff
@@ -129,3 +102,83 @@ def resolve_update_dataset(obj, info, input):
...
+        :returns: Record dictionary
+    '''
+
+    session = info.context['dbsession']
+    record = session.query(Record).filter(Record.id == input["id"]).first()
+
+    all_entries = input.pop('entry', [])
+    for entry in all_entries:
+        entry_input = {
+            "category": entry["category"],
```
jnu: just checking, it looks like the `EntryInput` in the schema allows for an ID field that's getting ignored here - are we planning on using that to allow for updating existing entries? (in case someone wants to edit an existing count)

api/mutations.py
nmanduj1: it works! https://i.pinimg.com/originals/ca/38/b8/ca38b8f7e096956f258d406a7820c7ba.gif  thank u! i will update throughout

api/mutations.py
nmanduj1: yup, youre right! I made the mod.

api/mutations.py
jnu: woohoo!!! glad to hear it! and btw nice job coming up with a workaround, and also triaging to not let this be a blocker!

for reference (if it's useful to know how i suggested this!) my thought process was:
1) maybe ariadne didn't bother to convert nested dictionaries at all, so i checked the docs. docs didn't state either way, so
2) i checked the [codez](https://github.com/mirumee/ariadne/blob/04ed6d1e446aa678c5081b73bfdf344ce52b84cd/ariadne/utils.py#L51) and confirmed they _did_ account for nested dicts (through a nice lil recursive func!), which, based on past experience, led me to think:
3) the decorators were probably just in the wrong sequence, bc i've been bitten by that many a time in the past

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+@convert_kwargs_to_snake_case
+def resolve_create_record(obj, info, input):
+    '''GraphQL mutation to create a Record.
+        :param input: params for new Record
+        :returns: Record dictionary
+    '''
+
+    session = info.context['dbsession']
+
+    record_input = {
```
jnu: why not just use `input` here instead of copying the values?

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+        "dataset_id": input["dataset_id"],
+        "publication_date": input["publication_date"]
+    }
+
+    record = Record(**record_input)
+    session.add(record)
+    session.commit()
+
+    all_entries = input.pop('entries', [])
```
jnu, line 103: one downside of `pop` is that it actually changes `input`, even though all you're trying to do is access the value `entries`. i'm not sure if it matters for ariadne or not if you change the `input`, but it's a best practice to avoid mutating your input arguments unless that's an intentional part of your design (e.g., for optimization or certain "builder" patterns)

here it looks like `input.get('entries', [])` would work nicely!

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+    for entry in all_entries:
+        n_entry = Entry()
```
jnu: looks like you could do this in one or two lines, like

```py
session.merge(Entry(record=record, **entry))
```

maybe im missing something tho?

jnu: also hm, from our call this morning i got the impression we decided _not_ to follow the `upsert` pattern? under what circumstances would we be updating existing Entries here? (`session.merge` does an upsert, so just checking!)

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+           setattr(n_entry, param, single_entry[param])
+        merged_object = session.merge(n_entry)
+
+    for param in input:
+        setattr(record, param, input[param])
+
+    session.add(record)
+    session.commit()
+
+    updated_record = session.query(Record).filter(Record.id == input["id"]).first()
```
jnu: fyi here and elsewhere, sqlalchemy actually takes care of this for you! after you call `session.commit` the objects in your session will be updated with the correct state, so you can just return `record` here. 

if you use `echo=True` in the create_engine statement you should be able to see all the SQL that gets generated under the hood. there should be a "select" statement issued to update the objects on commit -- this might be interested to do to understand some of the magic thats going on!

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+@mutation.field("updateRecord")
+@convert_kwargs_to_snake_case
+def resolve_update_record(obj, info, input):
+    '''GraphQL mutation to update a Record.
+        :param input: params to update Record
+        :returns: Record dictionary
+    '''
+
+    session = info.context['dbsession']
+    record = session.query(Record).filter(Record.id == input["id"]).first()
```
jnu: fyi here and elsewhere, sqlalchemy provides a shortcut for selecting a single item on the primary key: `session.query(Record).get(input['id'])`

api/mutations.py
```diff
@@ -129,3 +102,76 @@ def resolve_update_dataset(obj, info, input):
...
+    for single_entry in all_entries:
+        n_entry = Entry()
+        for param in single_entry:
+           setattr(n_entry, param, single_entry[param])
+        merged_object = session.merge(n_entry)
+
+    for param in input:
+        setattr(record, param, input[param])
+
+    session.add(record)
```
jnu, line 135: should this be a `merge` since we're updating?

api/mutations.py
```diff
@@ -1,8 +1,9 @@
 import datetime
 import uuid
+import re
```
jnu: looks like this can be removed now!

api/database.py
```diff
@@ -219,16 +219,16 @@ class Record(Base):
...
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
 
     category = Column(String(255), nullable=False)
     category_value = Column(String(255), nullable=False)
     count = Column(Integer, nullable=False)
-    record = relationship('Record', back_populates='data')
-    record_id = Column(GUID, ForeignKey('record.id'), index=True)
+    record = relationship('Record', back_populates='entries')
+    record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
```
jnu, line 233: does this mean we're not going to do the "soft deletes" we talked about last week on the call? (we had said we wouldn't actually delete records by design in case they needed to be retrieved -- this is a common database practice as insurance against mistakes)

jnu (commented):
i think this is looking really good! i made a few comments that might help clean things up a little more! nice work!!

api/mutations.py
```diff
@@ -129,3 +85,66 @@ def resolve_update_dataset(obj, info, input):
...
+        :returns: Record dictionary
+    '''
+
+    session = info.context['dbsession']
+
+    entries = []
+    all_entries = input.pop('entries', [])
+
+    for entry in all_entries:
+        new_entry = Entry(**entry)
```
jnu: this is nice! in python also as we've talked about it's usually better (quicker for humans _and_ machines!) to write the for-loop on one line, in a list comprehension
```py
entries = [Entry(**entry) for entry in all_entries]
```

api/database.py
seerocode: I imagine a soft delete means you have to update the deleted field for a record. But does this mean when you're filtering or querying for all records you'll have to remember to not include "deleted" records every time? Or maybe there's a better way.

seerocode: Also if the language in the UI specifies that a delete action is permanent, do we need to to do a soft delete? I think at that point the user has double confirmed and registered that deleting was a deliberate action on their part and not a mistake.

nmanduj1: im sure there's a better way- (insert pleading face emoji here)  I'm still working on cleaning up the mutations (specifically the merge we were talking about on slack) but have this on my list of things to look into.

api/mutations.py
nmanduj1: Q: Can you use list comprehension in python if youre doing more than 1 thing?

api/mutations.py
nmanduj1: thanks for this!

api/mutations.py
```diff
@@ -12,120 +11,132 @@
...
 
-    dataset_input = {
-        "name": input["name"],
-        "description": input["description"],
-        "program_id": input["programId"],
-        "inputter_id": input["inputterId"],
-    }
+    for tag in all_tags:
+        new_tag = Tag(**tag)
+        tags.append(new_tag)
```
jnu: forget whether i made this comment already but you will almost always see this written as a list comprehension instead of a for-loop:
```py
tags = [Tag(**tag) for tag in all_tags]
```

this is the best practice in python because it's 1) less code for humans to read, and 2) significantly faster for the computer to run (tho with these tags lists that won't make a huge diff)

api/mutations.py
```diff
@@ -12,120 +11,132 @@
...
-    '''GraphQL mutation to delete a dataset.
-        :param id: UUID of Dataset to be deleted
-        :returns: UUID of deleted Dataset
+    '''GraphQL mutation to soft delete a Dataset.
+        :param id: UUID of Dataset to be soft deleted
+        :returns: UUID of soft deleted Dataset
     '''
     session = info.context['dbsession']
-    session.query(Dataset).filter(Dataset.id == id).delete()
+    session.query(Dataset).filter(Dataset.id == id).update({'deleted':datetime.datetime.now()})
```
jnu, line 52: this is fine for now but we should talk at some point about adding a method to the Dataset class to handle this soft delete for us!

api/database.py
```diff
@@ -209,8 +209,10 @@ class Record(Base):
     dataset = relationship('Dataset', back_populates='records')
     dataset_id = Column(GUID, ForeignKey(
         'dataset.id'), nullable=False, index=True)
-    publication_date = Column(DateTime, index=True, unique=True)
-    data = relationship('RecordData')
+    publication_date = Column(DateTime, index=True)
+    entries = relationship('Entry', back_populates='record')
+    __table_args__ = (UniqueConstraint('dataset_id', 'publication_date', name='uix_dataset_id_publication_date'),
```
seerocode, line 214: Sweeeeet!!!

api/mutations.py
```diff
@@ -12,120 +11,132 @@
...
+    tags = []
+    all_tags = input.pop('tags', [])
 
-    # iterating over tags to check if the included tag exists or should be created.
-    for tag in tags:
-        existing_tag = session.query(Tag).filter(Tag.name == tag["name"].capitalize().strip()).one_or_none()
+    for tag in all_tags:
+        incoming_tag = Tag(**tag)
+        tags.append(incoming_tag)
+        merged_tag = session.merge(incoming_tag)
```
jnu: looks like the `merged_tag` variable is unused so could clean that up by removing the assignment!

api/mutations.py
```diff
@@ -12,120 +11,132 @@
...
 
-    # removing tags from input, and returning the contents of tags OR an empty array if tags was not found.
-    tags = input.pop('tags', [])
+    tags = []
+    all_tags = input.pop('tags', [])
 
-    # iterating over tags to check if the included tag exists or should be created.
-    for tag in tags:
-        existing_tag = session.query(Tag).filter(Tag.name == tag["name"].capitalize().strip()).one_or_none()
+    for tag in all_tags:
```
jnu, line 75: same comment here about list comprehension!

api/mutations.py
jnu: (o hey, i _did_ make this comment already 😬 !! sorry!!!) 

no you can't, you can only do one operation in a list comprehension, and that operation cannot be an assignment -- these groundrules are why the comprehensions can be so much faster than regular loops. but they were invented to improve the exact thing you're doing here - iterate over a list, call some function, and append to a new list - bc that's so common!

api/mutations.py
jnu: looks like this comment is still relevant?

api/queries.py
```diff
@@ -25,10 +25,19 @@ def resolve_user(obj, info, id):
 def resolve_dataset(obj, info, id):
     '''GraphQL query to find a dataset based on dataset ID.
         :param id: Id for the dataset to be fetched
-        :returns: Dataset dictionary
+        :returns: Dataset dictionary OR None if Dataset was soft-deleted
```
jnu, line 28: woohoo thanks for keeping these docs up-to-date!

api/queries.py
```diff
@@ -25,10 +25,19 @@ def resolve_user(obj, info, id):
...
+    return dataset
 
+@query.field("record")
+def resolve_record(obj, info, id):
+    '''GraphQL query to find a Record based on Record ID.
+        :param id: Id for the Record to be fetched 
+        :returns: Record dictionary OR None if Record was soft-deleted
+    '''
+    session = info.context['dbsession']
+    record = session.query(Record).filter(Record.id == id, "deleted" == None).first()
```
jnu: just flagging for the future that we'll want to put this in a method on the Record class (or something like that) so we don't have to repeat it in every single resolver. don't bother doing this now!

jnu (approved):
this looks good! i made a few minor suggestions that would be nice to get in before landing. overall nice work!!

PR #50, opened 2021-05-04
Docker compose

jnu:
In this PR:
 - [x] Split dockerfiles into `client/` and `api/`
 - [x] Add a docker-compose config, with `postgres` and `nginx` in separated containers
 - [x] Init script to magically create database when starting from scratch

Left to do before merging:
 - [x] Test compose networking (looks like maybe an issue with /graphql currently)
 - [x] Move DB password to `secret`
 - [x] Set app secret from a `secret`


api/docker-api-init.sh
```diff
@@ -0,0 +1,30 @@
...
+set -x
+
+# Create the passfile in the format `host:port:database:user:password`
+echo "$db_host:5432:$db_name:postgres" > .pgpass.tmp
+# Avoid echoing the password to the screen here
+paste -d':' .pgpass.tmp /run/secrets/db_pw > .pgpass
+chmod 600 .pgpass
+rm .pgpass.tmp
+
+until PGPASSFILE=.pgpass psql -h "$db_host" -U "postgres" -c '\q' "$db_name"; do
```
seerocode, line 13: What does until do? The internet isn't answering me.

api/docker-api-init.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Run a smoke test to see if the db needs initialization
+PGPASSFILE=.pgpass psql -h "$db_host" -U "postgres" -c 'select * from organization;' "$db_name"
+st=$?
+if [ $st -eq 0 ]; then
+    >&2 echo "Database already initialized"
+else
+    >&2 echo "Database hasn't been initialized yet, doing that now ..."
+    if [ "$RT_DUMMY_DATA" = "1" ]; then
+        python database.py --dummy-data
```
seerocode, line 28: Is this adding dummy data for a test environment?

README.md
```diff
@@ -13,30 +13,17 @@ Frontend
 Backend API
 [/api](https://github.com/stanford-policylab/bbc-50-50/tree/main/api)
 
-### Prerequisites
+### Docker compose
```
seerocode: This makes me so happy because this straggler documentation was eating at me.

api/docker-api-init.sh
jnu: which part? there's a lot going on!!

`PGPASSFILE` is a way to securely enter the password for `psql` which is the postgres CLI

`-h` and -U` specify the host and username where postgres is running

`-c '\q'` executes the command `\q` in the postgres CLI, which just means "quit"

and `db_name` is the database where our data live.

the tl;dr for that command is to check "is postgres running and is our database ready." if not, we use `until` in bash to loop until it _is_ ready.

docker-compose will start our db/api/nginx/etc all at the same time, and if you try to run this script before postgres is actually ready to receive queries it will fail.

i can add a short comment to note this!

api/docker-api-init.sh
jnu: yep, it's optional, you can control this with an environment variable passed to the docker-compose config

api/docker-api-init.sh
seerocode: Wooh I don't even know where to begin to search for this on the internet. Thanks for explaining it!

jnu: the PGPASSFILE in particular is important -- when writing scripts like this you want to be careful never to specify passwords on the commandline. usually the tools you're using will have ways to input a password that keep them hidden in files and never printed to the screen. the PGPASSFILE is kind of a wonky way to do it, but it works!

seerocode (approved):
🚢

PR #52, opened 2021-05-06
[ch205] Add sort for records table

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/205

### Changes:
- Added sorter method to table that sorts the publication date column in ascending or descending order (default sort = descending)
- Fixed the CSS for the layout pane so the footer doesn't overlap it and cut off the table content
- Removed the limit of 6 records that was put in place as a workaround for the issue above

### DIFF
--- a/client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
+++ b/client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
@@ -81,6 +81,7 @@ const DatasetDetailsRecordsTable = ({
         dayjs(dateB.publicationDate).unix(),
       fixed: true,
       width: 100,
+      render: (date: string) => dayjs(date).format("YYYY-MM-DD"),
     },
     {
       title: "Men",
--- a/client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
+++ b/client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
@@ -1,12 +1,14 @@
-import { Button, message, Popconfirm, Space, Table } from "antd";
+import { Button, Popconfirm, Space, Table } from "antd";
 import React from "react";
 import "./DatasetDetailsRecordsTable.css";
 import { GetDataset_dataset_records } from "../../__generated__/GetDataset";
 import { GET_DATASET } from "../../__queries__/GetDataset.gql";
+import { DELETE_RECORD } from "../../__mutations__/DeleteRecord.gql";
 import { useTranslation } from "react-i18next";
 import { useMutation } from "@apollo/client";
-import { DELETE_RECORD } from "../../__mutations__/DeleteRecord.gql";
 import { useHistory } from "react-router-dom";
+import { messageError, messageInfo, messageSuccess } from "../Message";
+import dayjs from "dayjs";
 
 interface DatasetRecordsTableProps {
   datasetId: string;
@@ -44,6 +46,12 @@ const DatasetDetailsRecordsTable = ({
     deleteRecord,
     { loading: deleteRecordLoader, error: deleteRecordError },
   ] = useMutation(DELETE_RECORD, {
+    onError: (error) => {
+      messageError(`${error.message}. Please try again later.`);
+    },
+    onCompleted: (deleted) => {
+      if (deleted) messageSuccess("Succesfully deleted record!");
+    }, // TODO: update cache instead of refetching
     refetchQueries: [
       {
         query: GET_DATASET,
@@ -52,27 +60,25 @@ const DatasetDetailsRecordsTable = ({
     ],
   });
 
-  const confirmDelete = async (recordId: string) => {
-    await deleteRecord({ variables: { id: recordId } })
-      .then(() =>
-        deleteRecordLoader
-          ? message.loading("Deleting record...")
-          : message.success("Succesfully deleted record!")
-      )
-      .catch((error) =>
-        message.error(`${error.message}. Please try again later.`)
-      );
+  const confirmDelete = (recordId: string) => {
+    deleteRecord({ variables: { id: recordId } });
   };
 
   const cancelDelete = () => {
-    message.info("Delete cancelled");
+    messageInfo("Delete cancelled");
   };
 
+  // TODO: render columns dynamically by dataset category types
   const columns = [
     {
       title: "Date",
       dataIndex: "publicationDate",
       key: "id",
+      sorter: (dateA: any, dateB: any) =>
+        dayjs(dateA.publicationDate).unix() -
+        dayjs(dateB.publicationDate).unix(),
+      fixed: true,
+      width: 100,
     },
     {
       title: "Men",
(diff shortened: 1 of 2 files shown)


client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -52,27 +60,25 @@ const DatasetDetailsRecordsTable = ({
...
+  // TODO: render columns dynamically by dataset category types
   const columns = [
     {
       title: "Date",
       dataIndex: "publicationDate",
       key: "id",
+      sorter: (dateA: any, dateB: any) =>
+        dayjs(dateA.publicationDate).unix() -
+        dayjs(dateB.publicationDate).unix(),
+      fixed: true,
```
seerocode, line 87: Fixes the date column to the left side so on scroll to the right, they can still see the dates for each record.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -52,27 +61,27 @@ const DatasetDetailsRecordsTable = ({
...
+    messageInfo("Delete cancelled");
   };
 
+  // TODO: render columns dynamically by dataset category types
   const columns = [
     {
       title: "Date",
       dataIndex: "publicationDate",
       key: "id",
+      defaultSortOrder: "descend" as SortOrder,
```
jnu: using `as` in typescript is a sign that your typechecking is not as strong as it could be. in this case, typescript doesn't know what `columns` is supposed to be when you define it, so it'll let you write anything -- and only later on when you reference `columns` does it try to reconcile that freeform type with the constraint AntD intended. but at that point it's a little too late to get it totally correct. one effect is what you saw here that it's not able to guarantee the string enum field SortOrder. another bad effect is often to assume `any` for generics.

in this case you should give `columns` an explicit type, so that you don't have to do `as SortOrder` here, and so the column object gets verified as much as possible as soon as you define it (vs when you use it).

```ts
const columns: ColumnType<OurRecordType>[] = [ ... ]
```

Since ColumnType is generic, the antd package could be adding type constraints to the `key`, `dataIndex`, and other fields as well, like this for example:

```ts
type ColumnType<T extends {}> = {
  key: keyof T;
  dataIndex: keyof T;
};
```

so that if you reference a field that doesn't exist on your record it'll catch that for you.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -146,7 +155,7 @@ const DatasetDetailsRecordsTable = ({
       scroll={{ x: 1000 }}
       sticky
       title={() => t("datasetRecordsTableTitle", { title: "Records" })}
-      pagination={{ pageSize: 6, hideOnSinglePage: true }}
```
jnu, line 149: just curious what the pageSize is now?

jnu (approved):
looks great! gave ya one tip about making more of the typechecking, but this is awesome. nice work ro!

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: I'm almost in the air and sneaking in a reply! I believe it's 20 or so records? I haven't counted and can't remember the default. It scrolls down quite a bit before it generates a second, etc. page.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: I think I had a type on the columns before and I didn't need to do any casting. I looked up this problem in the ant issues and it looks like a typescript problem that they don't have documentation for. 

I'll try your suggestion if I can't fish out the commit I had with the column type!

jnu: this isn't an antd problem or a typescript problem. (typescript is actually _correct_ in calling out your type!) this situation comes up whenever you define a mutable string without an explicit type and try to use it where a string enum is asked for. simpler example:

```ts
// FooType is a literal string enum type
type FooType = 'foo';

// this variable can't be inferred by TS to be FooType because it's
// mutable; your code could change it to anything. So TS infers that
// it is a string.
let foo = 'foo';

// This function requires the FooType enum as input
const f = (input: FooType) => console.log(input);

// This throws an error because TS can't know through static analysis if
// `foo` still conforms to FooType, since it can be modified at runtime.
f(foo);
```

the solution is to give `foo` an explicit type with `let foo: FooType = 'foo'`, or to make `foo` immutable as a `const`. the same thing is going on with your `columns.defaultSortOrder`.

this issue was discussed in a long thread when string enums were first implemented in typescript a few years ago -- i can try to find that if you're interested.

client/src/components/DataEntry/DataEntryMessageHandler.tsx
```diff
@@ -26,5 +26,6 @@ export const formMessageHandler = ({
     return messageSuccess({
       content: `Success! Your ${messageText} record has been saved.`,
       key: messageKey,
+      duration: 3,
```
nmanduj1: Just a clarifying q re: duration and accessibility-  I know we talked about accessibility for these messages and am wondering if this is enough time to be read ?

nmanduj1 (approved):

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
jnu: here's the PR that discusses the expected behavior of literal types, explains the theory behind the design choices, and links to a few related discussion threads around it: https://github.com/microsoft/TypeScript/pull/10676

seerocode: Thanks! I just needed to add the same type used for data as a `ColumnsType` for the columns array and antd checks the type for the rest. I had this before but can't remember why or when I removed it.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: The default is actually 10 records: 
image

client/src/components/DataEntry/DataEntryMessageHandler.tsx
seerocode: For this toast message, the loading toast and the success toast are connected via a key. Because the same key is used on each success message, you don't really know which date it was saved for. 

Initially, I thought about the length of the message which is at most 8 words and what the message was trying to convey. For an error message or messages with great length, 6 seconds is the minimum. Here the context is the user is informed that their action was successful which includes a green check mark as well which made me think 3 seconds was sufficient to read the message.

In writing this out, I realized there is a different way to handle this that would be more appropriate and will move to a different PR since it's not relevant for this story.

I dropped this commit in the meantime.

PR #53, opened 2021-05-07
Ch196/create tests dataset

nmanduj1:


api/schema.graphql
```diff
@@ -142,13 +142,13 @@ type Query {
   user(id: ID!): User
 
   # Retrieve a single dataset.
-  dataset(id: ID!): Dataset!
```
seerocode, line 145: I'm wondering why this was changed. Does a non-nullable return create an issue?

nmanduj1:
I am folding this PR into https://github.com/stanford-policylab/bbc-50-50/pull/54 for simpler merging.

PR #54, opened 2021-05-07
Ch215/create tests records

nmanduj1:


api/queries.py
```diff
@@ -40,4 +41,9 @@ def resolve_record(obj, info, id):
     '''
     session = info.context['dbsession']
     record = session.query(Record).filter(Record.id == id, Record.deleted == None).first()
-    return record
\ No newline at end of file
+    return record
+
+# This handles converting publicationDate to string- Ariadne is not converting automatically and json does not have DateTime data type
+@record.field("publicationDate")
```
seerocode: I think you can also add a global function for the date scalar. Like here: https://ariadne.readthedocs.io/en/0.2.0/scalars.html

That way any time date is used in a type it won't error.

jnu: +1 scalars are the way to go here!

it's also best to use a standardized date format like ISO-8601 (you can call `.isoformat()` to get this). python also does not automatically include timezones in dates like JS does, so we should make sure the object's `tzinfo` is getting populated correctly.

api/app_test.py
```diff
@@ -86,6 +87,257 @@ def test_query_user(self):
...
 
+    def test_query_dataset(self):
+        success, result = self.run_graphql_query({
+            "operationName": "QueryDataset",
+            "query": """
+                query QueryDataset($id: ID!) {
+                   dataset(id: $id) {
+                        id
+                        name
+                        description
```
jnu, line 140: maybe you could add a few more fields to this query? more fields should be pretty quick to add and will give us more coverage. to your point last week about knowing "how much to test" - you should at least make sure that each of the fields we're able to fetch from `dataset` work.

you don't necessarily need to get the maximal nested data that the children allow here, as long as you're going to cover that in other tests. but overall, every field that can be accessed through a graphql should be covered somewhere in our tests.

api/app_test.py
```diff
@@ -86,6 +87,257 @@ def test_query_user(self):
...
+                    "tags": [{"name": "Europe", "description": "i am europe", "tagType": "location"}]
+                } 
+            },
+        })
+
+        self.assertTrue(success)
+        self.assertEqual(result, {
+            "data": {
+                "createDataset": {
+                    "name": "Happy Hour",
```
jnu, line 240: it would be reasonable to check somehow that you got a uuid back as the ID as well, since this is a critical behavior of the `create` operation that's not necessarily a given.

api/app_test.py
```diff
@@ -86,6 +87,257 @@ def test_query_user(self):
...
+                    "tags": [{"name": "Asia", "description": "i am asia", "tagType": "location"}]
+                } 
+            },
+        })
+
+        self.assertTrue(success)
+        self.assertEqual(result, {
+            "data": {
+                "updateDataset": {
+                    "name": "Tea Time",
```
jnu, line 281: could you return and validate the ID field here as well, to make sure it was the same one that was passed in? similar to create, this is a critical behavior of the update mutation!

api/app_test.py
```diff
@@ -86,6 +87,257 @@ def test_query_user(self):
...
+                    "description": "Tea time programming",
+                    "program": {"name": "BBC News"},
+                    "tags": [{"name": "News"}, {"name": "Asia"}]
+                },
+            },
+        })    
+
+    def test_delete_dataset(self):
```
jnu, line 289: we should make assertions about the soft / cascading behavior of this delete in the database, since these are the trickiest behaviors of our api in this mutation.

to do this, you can use the test's db session and import the relevant objects from `database.py` to make assertions on the results of queries against the database to check whether the relevant rows (dataset, record, and entry) exist in the deleted state after this mutation

api/app_test.py
```diff
@@ -86,6 +87,257 @@ def test_query_user(self):
...
+                "updateRecord": {
+                    "publicationDate": "2020-12-25 00:00:00",
+                    "dataset": {"id": "96336531-9245-405f-bd28-5b4b12ea3798", "name": "12PM - 4PM"},
+                    "entries": [{"categoryValue": "trans", "count": 10}]
+                },
+            },
+        })   
+
+    def test_delete_record(self):
+        success, result = self.run_graphql_query({
```
jnu, line 449: similar to the `delete_dataset`, we should validate that the proper hard delete with cascading took place after this mutation in the sqlite database!

api/schema.graphql
```diff
@@ -142,13 +142,13 @@ type Query {
...
   # Retrieve a single dataset.
-  dataset(id: ID!): Dataset!
+  dataset(id: ID!): Dataset
 
   # Retrieve a single Record.
   record(id: ID!): Record
 
   # Retriever a single program.
-  program(id: ID!): Program!
+  program(id: ID!): Program
```
jnu: just checking, have you and @seerocode agreed on the expected behavior for missing objects in queries? what's the use case for returning `null` values that aren't errors from these queries?

api/queries.py
jnu: also just wanna note that i know you two were discussing how to represent a "Date" type that doesn't include a "Time" component. most languages / databases don't have a separate type for this so you use midnight at UTC time as a convention -- however if you wanted to you could add a custom DateWithoutTime scalar to the graphql api and format it as `%Y-%m-%d`!

jnu (changes requested):
great start, this is v close to being ready to merge! i added a couple notes to get more use out of the tests, and i think @seerocode is right about the date scalar! 

i'm really excited to see more tests in here!! great work!!!

api/queries.py
seerocode: I'm pretty sure you can also change `DateTime` in the DB back to `Date` and leave the scalar in the schema as is. If you're creating a custom scalar function like in the Ariadne docs, you can probably do something like:

```python
import datetime

date_scalar = ScalarType("Date")


@date_scalar.serializer
def serialize_date(value: datetime.date) -> str:
    return value.isoformat()

@date_scalar.value_parser
def parse_date(value: str) -> datetime.date:
    return datetime.datetime.strptime(value, "%Y/%m/%d").date()
```

Probably don't even need the value parser.

This way the DB saves Date as a Date without time for publicationdate and maybe create a different DateTime scalar in the schema down the road if we need it for any other date stuff that we need the time and timezone for.

jnu: it doesn't matter much to me what you choose, just watch out that JS does _not_ have a Date type that does not include time, so just make sure you're handling that correctly on the frontend!

nmanduj1: Thanks for pointing out the scalar! I updated and it cleaned the code up nicely. I am leaving the docs relevant to our ariadne version (0.12.0) here for posterity: https://ariadnegraphql.org/docs/0.12.0/scalars Also for posterity- we decided to use the midnight at UTC convention.

Question: I can't seem to find a way to dynamically set the object's timezones - @jnu do we just assume a timezone and set all datetime objects to the same? I am seeing a way to handle this statically using the pytz library but I'm not sure what the convention is to dynamically handle setting local timezones in python.

api/app_test.py
nmanduj1: I added more fields throughout! Hopefully the additional coverage will help us avoid 🐛 🐛 🐛

api/app_test.py
nmanduj1: I added a "is_valid_uuid" method to handle this check throughout-

api/app_test.py
nmanduj1: Thank you for pointing this out! I added a check for this-

api/app_test.py
nmanduj1: I added tests to ensure the hard delete is working as expected. However I am running into a problem with sqllite- it does not like our db.py cascade delete flag. I am able to confirm cascading hard deletes are behaving as expected through the graphql playground tho. Any thoughts on how I should update the flag to get sqllite to play nice? Or should we use a postgres db to confirm functionality instead?

api/schema.graphql
nmanduj1: I updated this! Thank you both @jnu and @seerocode for clarifying!

api/app_test.py
jnu: nice!!

api/app_test.py
jnu: this testing is phenomenal, thank you so much for adding!!!

for future reference: the issue with cascades is that sqlite doesn't have foreign keys enabled by default, which means that the cascades wouldn't work. the solution is to enable foreign keys when setting up the sqlite db, which can be done with the query:
```
PRAGMA foreign_keys = ON
```

I added this to the test setup function and it works now!

api/app_test.py
jnu: this looks great, thank you so much for adding!!!

api/app_test.py
jnu: this is so thorough i love it so much!! nice work!!

api/queries.py
jnu: you shouldn't ever have to assume a timezone, ideally every timestamp that gets passed around has timezone information included with it

jnu: fyi i've added the module python-dateutil to help with parsing here. it's a lot more flexible than the built-in `datetime.fromisoformat`

also wanna emphasize that for scalars the `value_parser` function is _required_, because it's impossible to make a graphql request with a "date object" in it -- the requests are sent to the graphql API as JSON, and JSON does not have a date type, so we have to use either a string or a number instead. i've added this deserializer function!

jnu (approved):
i made a couple of small changes to finish this up. great work!!

api/queries.py
nmanduj1: thanks for pointing this out, @jnu. Didn't think I needed this since postgres is smart enough to parse the string into a datetime on its own. Clearly not the case for sqllite! Great learning opportunity for me here, thank you!

PR #55, opened 2021-05-12
[ch164] Breadcrumbs in dataset overview page

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/164

image
image
image
image

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -1,12 +1,12 @@
 import React, { Suspense } from "react";
-import { Switch, Route, BrowserRouter, Redirect } from "react-router-dom";
+import { BrowserRouter } from "react-router-dom";
 import { Layout } from "antd";
 import AppHeader from "./layout/AppHeader";
 import { AppSidebar } from "./layout/AppSidebar";
-import { Home } from "./components/Home/Home";
-import { DatasetDetails } from "./components/DatasetDetails/DatasetDetails";
-import { DataEntry } from "./components/DataEntry/DataEntry";
 import "./App.css";
+import ROUTES from "./router/routes";
+import BreadCrumb from "./components/Breadcrumb/Breadcrumbs";
+import { RenderRoutes } from "./router/Router";
 
 const { Footer, Content } = Layout;
 
@@ -20,31 +20,8 @@ function App() {
             <AppSidebar />
             <Layout style={{ padding: "0 24px 24px" }}>
               <Content className="site-layout-background">
-                <Switch>
-                  <Route exact path="/" component={Home} />
-                  <Route
-                    exact
-                    path="/dataset/:datasetId/details"
-                    component={DatasetDetails}
-                  />
-                  <Route
-                    key="add-entry"
-                    exact
-                    path="/dataset/:datasetId/entry"
-                    component={DataEntry}
-                  />
-                  <Route
-                    key="edit-entry"
-                    exact
-                    path="/dataset/:datasetId/entry/edit/:recordId"
-                    component={DataEntry}
-                  />
-                  <Redirect
-                    exact
-                    from="/dataset/:datasetId/entry/reload"
-                    to="/dataset/:datasetId/entry"
-                  />
-                </Switch>
+                <BreadCrumb />
+                <RenderRoutes routes={ROUTES} />
               </Content>
               <Footer style={{ textAlign: "center" }}>Footer content</Footer>
             </Layout>
--- a/client/src/components/Breadcrumb/Breadcrumbs.tsx
+++ b/client/src/components/Breadcrumb/Breadcrumbs.tsx
@@ -1,0 +1,27 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import { Breadcrumb } from "antd";
+import ROUTES from "../../router/routes";
+import useBreadcrumbs from "use-react-router-breadcrumbs";
+
+const BreadCrumb = () => {
+  /**
+   * DisableDefaults disables all routes in a path
+   * that have not been set up in the router config
+   * e.g. "/dataset/:id" will only create a breacrumb for
+   * Home / Dataset Details and not Home / Dataset / Dataset Details
+   * because a route for "/dataset" does not exist
+   */
+  const breadcrumbs = useBreadcrumbs(ROUTES, { disableDefaults: true });
+  return (
+    <Breadcrumb>
+      {breadcrumbs.map(({ breadcrumb, match }) => (
+        <Breadcrumb.Item key={match.url}>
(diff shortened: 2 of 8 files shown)


client/src/components/Message.tsx
```diff
@@ -24,22 +24,24 @@ type Message = "info" | "success" | "error" | "warning" | "loading";
...
-};
+const createMessage =
+  (type: Message) =>
+  (
+    content: JointContent,
+    duration?: number | (() => void) | undefined,
+    onClose?: ConfigOnClose | undefined
+  ) => {
+    /* cast content to object */
+    const contentPropObject = content as Record<string, unknown>;
```
jnu: i think you can probably avoid this cast -- i think i see why you were probably seeing an error from typescript without it. on L39 you check if the content is a string, otherwise you try to spread it - it looks like this isn't *quite* enough because `ConfigContent` might also be a `ReactNode`, which will give you an error if you try to spread it. you'll have to duck-type both cases before typescript agrees you have ArgsProps, which should be spreadable.

client/src/components/Message.tsx
```diff
@@ -24,22 +24,24 @@ type Message = "info" | "success" | "error" | "warning" | "loading";
...
-      ? { style, content }
-      : { style, ...contentPropObject };
-
-  return message[type](newContent, duration, onClose);
-};
+const createMessage =
+  (type: Message) =>
+  (
+    content: JointContent,
+    duration?: number | (() => void) | undefined,
```
jnu: just fyi you don't need the `| undefined` if you mark the parameter as optional with the question mark!

client/src/router/Router.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { Route, Switch } from "react-router-dom";
+
+/**
+ * Flatten routes to render components for a subroute
+ */
+function flattenRoutes(route: any, agg: any = []) {
```
jnu: Can you use the real `route` object type here for the arguments? then you won't need the `any` on L18 either

client/src/router/Router.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  return agg;
+}
+
+/**
+ * Use this component for any new section of routes
+ * (any config object that has a "routes" property
+ */
+export function RenderRoutes({ routes }: any) {
```
jnu, line 29: same q here, there should be a reasonable type available for the `route` config object?

client/src/components/Breadcrumb/Breadcrumbs.tsx
```diff
@@ -0,0 +1,27 @@
...
+   * e.g. "/dataset/:id" will only create a breacrumb for
+   * Home / Dataset Details and not Home / Dataset / Dataset Details
+   * because a route for "/dataset" does not exist
+   */
+  const breadcrumbs = useBreadcrumbs(ROUTES, { disableDefaults: true });
+  return (
+    <Breadcrumb>
+      {breadcrumbs.map(({ breadcrumb, match }) => (
+        <Breadcrumb.Item key={match.url}>
+          <Link to={match.url || ""}>{breadcrumb}</Link>
```
jnu, line 20: huh, very cool! so this will properly generate the breadcrumb links? so if you're at `/dataset/1234/addrecord` (or whatever) it'll generate the link to `/dataset/1234/details` in the breadcrumb?

jnu (approved):
looks great! i love that you were able to figure out a clean way to generate these breadcrumbs! i left a couple smol comments about cleaning up the typings

client/src/components/Breadcrumb/Breadcrumbs.tsx
seerocode: Yep! I had changed the routes to remove any path that didn't have a view like `/dataset` but found I could disable this in the hook which is pretty cool. The hook uses a route matcher and `useLocation` hook like I had been doing originally to find the breadcrumb for the current path so using it in place of what I had originally worked a lot better.

client/src/router/Router.tsx
seerocode: I'll create an interface and extend this https://github.com/icd2k3/use-react-router-breadcrumbs/blob/master/src/index.tsx#L37 so it works.

client/src/components/Message.tsx
seerocode: Will change.

client/src/App.tsx
```diff
@@ -20,31 +20,8 @@ function App() {
...
-                    path="/dataset/:datasetId/entry/edit/:recordId"
-                    component={DataEntry}
-                  />
-                  <Redirect
-                    exact
-                    from="/dataset/:datasetId/entry/reload"
-                    to="/dataset/:datasetId/entry"
-                  />
-                </Switch>
+                <BreadCrumb />
```
nmanduj1, line 23: Whoa! Breadcrumbs cleaned this up a lot!

client/src/components/Breadcrumb/Breadcrumbs.tsx
nmanduj1: That it pretty cool!

client/src/components/Message.tsx
seerocode: What does `duck-type` mean?

client/src/App.tsx
seerocode: The `RenderRoutes` function does all this magic! I like putting the route configurations or having all the routes in one separate file as well.

nmanduj1 (approved):

client/src/components/Message.tsx
jnu: duck typing means determining the type of an object at runtime, instead of at build-time. you need to do this here because you have three different types that `content` could be: `string`, `ReactNode`, or `ArgsProps`, and you need to figure out which one you have while the function is running, bc you handle them differently. typescript is giving you an error because you've only checked the `string` case on L39 - that means on L41 you might be trying to spread a `ReactNode` with the `...` operator, which will throw an error.

so, on L39 you actually need to detect if you have a `ReactNode` too. unfortunately `typeof content` will give you `'object'` for either ReactNode or ArgsProps (because javascript is bad at types), so you need use another approach to check the type. usually in javascript you do this by "duck-typing," which means checking some characteristic of the object to see if it has some property or method (the name comes from: "if it walks like a duck, it must be a duck")

 I think React has a helper you can use: `React.isValidElement` which will cover both the `string` and `ReactNode` cases, so this can replace your `typeof content` check.

after you do this duck-typing properly you should be able to remove this `contentPropObject` on L35. this line worries me bc it might not actually be a valid `Record` object -- whenever `content` is a string or ReactNode, it's _not_ a Record and so this `as` type assertion is misinformation!

PR #56, opened 2021-05-13
[ch169] "Publication date" defaults to today for new data

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/169

Today's date was already default for adding new data.

### Changes:
- update data entry tests to expect the form date to be the correct format for add and edit entries
- add a mock date library for mocking dates in tests 
- parse datetime from record to date string (was broken before)

### DIFF
--- a/client/src/__tests__/DataEntry.test.tsx
+++ b/client/src/__tests__/DataEntry.test.tsx
@@ -10,6 +10,7 @@ import { autoMockedClient } from "../__mocks__/AutoMockProvider";
 import { GraphQLError } from "graphql";
 import { ApolloProvider } from "@apollo/client";
 import { DataEntry } from "../components/DataEntry/DataEntry";
+import MockDate from "mockdate";
 
 const history = createMemoryHistory();
 
@@ -82,10 +83,14 @@ test("should load and return error if dataset query fails", async () => {
   ).toBeInTheDocument();
 });
 
-test("should render add entry form when a record id does not exist in route params", async () => {
+test("should render add entry form with today's date when a record id does not exist in route params", async () => {
   (useParams as jest.Mock).mockReturnValue({
     datasetId: "5a8ee1d5-2b5a-49db-b466-68fe50a27cdb",
   });
+
+  // Sets a fixed date of Sunday, 14 June 2015 22:12:05.275
+  // (2015-06-14)
+  MockDate.set(1434319925275);
 
   const client = autoMockedClient();
 
@@ -106,7 +111,11 @@ test("should render add entry form when a record id does not exist in route para
   );
 
   expect(screen.getByRole("form"));
-  screen.getByLabelText("publicationDate");
+  expect(
+    screen.getByLabelText("publicationDate", {
+      selector: "input",
+    })
+  ).toHaveValue("2015-06-14");
 
   expect(screen.getAllByRole("textbox")).toHaveLength(6);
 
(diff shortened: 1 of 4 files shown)


client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -85,7 +85,7 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
   const [values, setValues] = useState<Entry[]>(entries);
 
   const [formPublicationDate, setFormPublicationDate] = useState<string>(
-    props.existingRecord?.record?.publicationDate ||
+    dayjs(props.existingRecord?.record?.publicationDate).format("YYYY-MM-DD") ||
```
seerocode: Converts datetime to date string when editing a record.

client/src/__tests__/DataEntry.test.tsx
```diff
@@ -82,11 +83,15 @@ test("should load and return error if dataset query fails", async () => {
...
 
-test("should render add entry form when a record id does not exist in route params", async () => {
+test("should render add entry form with today's date when a record id does not exist in route params", async () => {
   (useParams as jest.Mock).mockReturnValue({
     datasetId: "5a8ee1d5-2b5a-49db-b466-68fe50a27cdb",
   });
 
+  // Sets a fixed date of Sunday, 14 June 2015 22:12:05.275
+  // (2015-06-14)
+  MockDate.set(1434319925275);
```
jnu, line 93: very nice! i've been bitten so many times by forgetting to mock dates in tests!

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
jnu: im not sure i understand the use of `||` here -- what does calling `dayjs(undefined).format('YYYY-MM-DD')` do (i.e. if the existingRecord does not exist)?

it seems like it would either be an error, or redundant with `dayjs().format(''YYYY-MM-DD")`. (i'm guessing you tested this, which would indicate the latter! -- either way the `||` would only kick in if the first part returned `undefined`, which doesn't seem likely to me)

jnu (commented):
looks great! just one question about that `||`

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
seerocode: Good catch! It is redundant as dayjs sets the current date if the string passed to it is undefined. So this could be changed to `useState<string>(dayjs(props.existingRecord?.record?.publicationDate).format("YYYY-MM-DD"))` instead but you mentioned using localization.

seerocode: Since this is for the date picker to use the date provided (and not how the date will be displayed to the user), I removed the redundancy and left out localization.

client/src/__tests__/DataEntry.test.tsx
nmanduj1: Mockdate seems like a handy package. I'm not sure I follow here tho- what's the usefulness of bringing in mockdate if we are setting a fixed date on line 93 and then checking it on line 118? Wouldn't we want to mock a random date?

seerocode: @nmanduj1 a random date would make it difficult to make the assertion that the date rendered has a specific format (YYYY-MM-DD, in this instance). We could create a function to validate the format on any date provided but that seemed like extra code to maintain. By testing on a fixed date we can be confident that our tests won't flake due to a changing date in the test.

jnu (approved):
looks good to me!

PR #57, opened 2021-05-14
[ch220] Add "last updated" date to dataset for UI

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/220

Changes listed as tasks in the linked story.

image

### DIFF
--- a/api/queries.py
+++ b/api/queries.py
@@ -38,7 +38,7 @@ def resolve_dataset(obj, info, id):
 
 @dataset.field("lastUpdated")
 @convert_kwargs_to_snake_case
-def resolve_dataset_last_updated(dataset, info, *_):
+def resolve_dataset_last_updated(dataset, info):
     '''GraphQL query to find the date a dataset was last updated.
         :param dataset: Dataset object to filter by its ID
         :returns: Datetime scalar
--- a/api/queries.py
+++ b/api/queries.py
@@ -1,7 +1,11 @@
 from ariadne import convert_kwargs_to_snake_case, ObjectType
+from sqlalchemy.sql.expression import func
 from database import Dataset, User, Record
 
 query = ObjectType("Query")
+dataset = ObjectType("Dataset")
+
+queries = [query, dataset]
 
 '''GraphQL query to find a user based on user ID.
     :param obj: obj is a value returned by a parent resolver
@@ -31,6 +35,20 @@ def resolve_dataset(obj, info, id):
     dataset = session.query(Dataset).filter(Dataset.id == id, Dataset.deleted == None).first()
 
     return dataset
+
+@dataset.field("lastUpdated")
+@convert_kwargs_to_snake_case
+def resolve_dataset_last_updated(dataset, info, *_):
+    '''GraphQL query to find the date a dataset was last updated.
+        :param dataset: Dataset object to filter by its ID
+        :returns: Datetime scalar
+    '''
+    session = info.context['dbsession']
+
+    return session.query(func.max(Record.updated)).\
+            filter(Record.dataset_id == dataset.id, Record.deleted == None).\
+                scalar()
+
 
 @query.field("record")
 def resolve_record(obj, info, id):
--- a/api/schema.graphql
+++ b/api/schema.graphql
@@ -72,6 +72,7 @@ type Dataset {
   id: ID!
   name: String!
   description: String
+  lastUpdated: Date
   program: Program!
   records: [Record!]!
   inputter: User!
--- a/api/schema.graphql
+++ b/api/schema.graphql
@@ -72,7 +72,7 @@ type Dataset {
   id: ID!
   name: String!
   description: String
-  lastUpdated: Date
+  lastUpdated: DateTime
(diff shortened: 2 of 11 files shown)


client/src/components/Home/Home.tsx
```diff
@@ -78,7 +79,7 @@ const getTableData = (queryData: GetUser | undefined) => {
           id: dataset.id,
           team: program.name,
           dataset: dataset.name,
-          lastUpdated: "12/20/2020",
+          lastUpdated: dayjs(dataset.lastUpdated).format("YYYY-MM-DD"),
```
seerocode: Should we include the timestamp as well?

api/queries.py
```diff
@@ -32,6 +36,20 @@ def resolve_dataset(obj, info, id):
 
     return dataset
 
+@dataset.field("lastUpdated")
+@convert_kwargs_to_snake_case
+def resolve_dataset_last_updated(dataset, info, *_):
```
jnu: just curious why the `*_` is needed? does ariadne pass in some random arguments here or something?

api/schema.graphql
```diff
@@ -72,6 +72,7 @@ type Dataset {
   id: ID!
   name: String!
   description: String
+  lastUpdated: Date
```
jnu: just checking, is this the type that includes the time too or no?

client/src/components/Home/Home.tsx
jnu: i'd so no to start, maybe you can put it in the tooltip text or something.

what do you think about formatting the date with localization instead of y-m-d?

api/queries.py
```diff
@@ -32,6 +36,20 @@ def resolve_dataset(obj, info, id):
...
+    '''GraphQL query to find the date a dataset was last updated.
+        :param dataset: Dataset object to filter by its ID
+        :returns: Datetime scalar
+    '''
+    session = info.context['dbsession']
+
+    return session.query(func.max(Record.updated)).\
+            filter(Record.dataset_id == dataset.id, Record.deleted == None).\
```
jnu, line 49: is it possible that an `entry` was updated more recently than the record itself or no?

jnu (approved):
looks great to me! nice work!!

client/src/components/Home/Home.tsx
seerocode: @jnu do you mean like this? https://day.js.org/docs/en/display/format#localized-formats

api/queries.py
seerocode: I'm actually not sure. I didn't need these so maybe it's okay to leave it out of the signature.

api/schema.graphql
seerocode: Yeah. However, @nmanduj1 is changing this in her PR so when #54 is merged, I'll rebase to include the updated scalar for date time.

api/queries.py
seerocode: Since we are passing the whole Record object when we update an entry instead of just its entries, the Record itself gets updated too. Initially I had checked for the entries instead but once I realized this, I changed it to only check for the last updated record.

Also, not sure if we want to change how we update a record in the future and I know we had discussed this. If a specific field was updated and not the whole record, it makes sense to just update the entry instead of the entire record if we want to be completely correct about what was updated.

client/src/components/Home/Home.tsx
jnu: yep! i like to use `l` or `L` for dates that'll be shown to normal humans (maybe one of the longer-form ones too, if you want)! 

personally i think Y-M-D is the only logical format to use, but non-computer people have always told me i'm wrong!

api/queries.py
jnu: ok yep! python (unlike javascript) will complain loudly if the arguments are wrong. i'm pretty sure ariadne only needs the extra args here if your resolver itself takes arguments (like the `id` for dataset)

client/src/components/Home/Home.tsx
seerocode: I like year, month, day too. Maybe using `January 10, 2020` would be more readable for everyone unless UK reads it as `10 January, 2020`. 😟

api/queries.py
seerocode: Fixed!

client/src/components/Home/Home.tsx
seerocode: I decided to use the `ll` format here instead: 
image

seerocode:
Pending merge of #54

client/src/components/Home/Home.tsx
nmanduj1: That looks good to me!

api/queries.py
nmanduj1: Yea, that makes sense to me, Ro. Currently updating a record is the only way to update an Entry- I'm curious if we can come up with an intuitive UI to decouple Records from Entries to allow modification of only Entries in the future...

nmanduj1 (approved):

PR #58, opened 2021-05-17
Fix husky hook

seerocode:
The pre-commit husky hook wasn't working. Tried the suggestion [here](https://dev.to/devictoribero/how-to-use-husky-to-create-pre-commit-and-pre-push-hooks-4448) and in the [thread linked](https://github.com/typicode/husky/issues/445). Ended up upgrading to Husky v6 which uses a different configuration for hooks.

More details here and why husky hooks don't auto-install .git/ hooks anymore: 
- https://blog.typicode.com/husky-git-hooks-javascript-config/
- https://blog.typicode.com/husky-git-hooks-autoinstall/

Now it works!
image


jnu (approved):
interesting, thanks for figuring this out!

PR #59, opened 2021-05-17
Migrate from tslint to eslint

seerocode:
TSLint is now deprecated! See here: https://github.com/palantir/tslint

ESLint is now the standard linter for both Javascript and Typescript. This handy repo helped migrate to eslint from our tslint config: https://github.com/typescript-eslint/tslint-to-eslint-config

### Changes:
- Migrated to eslint and removed tslint config 
- Added strict rules for jsx-a11y
    - Fixed an issue that the strict rules discovered!
- Added prettier as an extension which should turn off rules that conflict with our prettier config (how I originally discovered tslint was deprecated) 
- Updated disable linter comments to use eslint instead of tslint
- Upgraded tslint and prettier


jnu (approved):
this is awesome, i love it! thank you so much for upgrading!

PR #62, opened 2021-05-24
[ch229] BE - Populate data entry form dynamically using advanced dataset metadata

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/229

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -244,8 +244,8 @@ class Entry(Base):
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
 
-    category = Column(String(255), nullable=False)
-    category_value = Column(String(255), nullable=False)
+    category_id = Column(GUID, ForeignKey('category.id'), index=True)
+    category = relationship('Category')
     count = Column(Integer, nullable=False)
     record = relationship('Record', back_populates='entries')
     record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
@@ -290,13 +290,6 @@ def create_dummy_data(session):
     ds2 = Dataset(id='96336531-9245-405f-bd28-5b4b12ea3798', name='12PM - 4PM', description='afternoon programming', inputter_id='cd7e6d44-4b4d-4d7a-8a67-31efffe53e77')
     program.datasets.append(ds1)
     program.datasets.append(ds2)
-
-    # datetime.strptime converts a string to a datetime object bc of SQLite DateTime limitation- must be explicit about format
-    record = Record(id='742b5971-eeb6-4f7a-8275-6111f2342bb4', dataset_id='b3e7d42d-2bb7-4e25-a4e1-b8d30f3f6e89', publication_date=datetime.strptime('2020-12-21 00:00:00', '%Y-%m-%d %H:%M:%S')) 
-    ds1.records.append(record)
-
-    entry = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', category='gender', category_value="female", count=8, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    record.entries.append(entry)
 
     tag = Tag(id='4a2142c0-5416-431d-b62f-0dbfe7574688', name='news', description='tag for all news programming',
             tag_type='news')
--- a/api/database.py
+++ b/api/database.py
@@ -198,6 +198,13 @@ class Category(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
     
+    @validates('category')
+    def capitalize_category(self, key, category):
+        return category.capitalize().strip()
+    
+    @validates('category_value')
+    def capitalize_category_value(self, key, category_value):
+        return category_value.capitalize().strip()
 
 class Dataset(Base):
     __tablename__ = 'dataset'
--- a/api/database.py
+++ b/api/database.py
@@ -189,14 +189,15 @@ class Category(Base):
     category = Column(String(255), nullable=False)
     category_value = Column(String(255), nullable=False)
 
-    targets = relationship('Target')
-
+    targets = relationship('Target', back_populates='category')
+    entries = relationship('Entry', back_populates='category')
+    
     created = Column(TIMESTAMP,
                      server_default=func.now(), nullable=False)
     updated = Column(TIMESTAMP,
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
-
+    
 
 class Dataset(Base):
     __tablename__ = 'dataset'
@@ -219,7 +220,6 @@ class Dataset(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
-
 class Record(Base):
     __tablename__ = 'record'
 
@@ -238,14 +238,13 @@ class Record(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
-
 class Entry(Base):
     __tablename__ = 'entry'
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
 
     category_id = Column(GUID, ForeignKey('category.id'), index=True)
-    category = relationship('Category')
+    category = relationship('Category', back_populates='entries')
     count = Column(Integer, nullable=False)
     record = relationship('Record', back_populates='entries')
     record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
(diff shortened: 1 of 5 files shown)


seerocode:
@nmanduj1 thank you for fixing the tests!!! 🙌🏽

nmanduj1:
np! I'm still working on fixing the create/update tests and associated resolvers- I will update when this is ready! ty!!!

api/database.py
```diff
@@ -331,6 +331,24 @@ def create_dummy_data(session):
     session.add(category_trans_men)
     session.add(category_gender_non_conforming)
 
+    # datetime.strptime converts a string to a datetime object bc of SQLite DateTime limitation- must be explicit about format
+    record = Record(id='742b5971-eeb6-4f7a-8275-6111f2342bb4', dataset_id='b3e7d42d-2bb7-4e25-a4e1-b8d30f3f6e89', publication_date=datetime.strptime('2020-12-21 00:00:00', '%Y-%m-%d %H:%M:%S')) 
+    ds1.records.append(record)
+
+    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', category_id='51349e29-290e-4398-a401-5bf7d04af75e', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4')
```
seerocode: @nmanduj1 I updated the record to include a complete set of entries for this dummy data.

seerocode:
I see the tests are passing now @nmanduj1 !! Is there anything left to include on the backend side?

nmanduj1:
@seerocode Yup, this is still wip from my end. The logic has to check if a Category already exists. Am I blocking your merge? You can merge if you'd like since the tests are passing and I can create a new branch to handle updating the resolvers.

seerocode:
@nmanduj1 no, not blocking. What do you mean if a Category already exists?

nmanduj1:
We don't want to create duplicate Categories so we have to look for the Categories by name before creating, the same way we do with Tags.

seerocode:
Oh, I see. It sounds like this is for creating a new category? @nmanduj1

nmanduj1:
Kinda, except the logic is nested in Creating a new Record or updating a Record. We used to have those resolvers handle updating and/or creating new Entries. Now that we've abstracted an Entry to have a Category, the logic in the createRecord and updateRecord resolvers has to be updated to allow updating or creating of the Category associated with each Entry associated with the Record. 

This is getting v complicated v quickly bc of the nested associated tables/ nested creates and updates. Honestly, it's becoming a bit overwhelming to keep track of everything. I'm starting to wonder if this is a good approach. @seerocode

nmanduj1:
@seerocode Our Entry table only has a count and two foreign keys. I feel like it's looking more like a join table than its own separate model.

seerocode:
@nmanduj1 maybe we can talk through it tomorrow? I don't know if we need to create a new category when creating a record. At least on the front end, when a record is created, we take the categories that already exist for a program (trickled down to the dataset) and set that category id that way but I don't think there's ever a time we would be creating a new category when updating/creating a record because we are only storing the ID that exists for a category already. This is what I'm seeing in the table now for entry:

image

There's the other question about the issue updating descriptions for categories that could also inform this.

nmanduj1:
Okedokes, that simplifies things. I will create the resolvers to handle CRUD for Categories. ty!! Also, I should create a new branch handling CRUD for Categories- I think this branch is able to be merged. Let me review/approve-

seerocode:
@nmanduj1 wait before you do, do you want to talk about the current Category table? I think you're right in that it's not the best set up because you might run into a similar same issue you described in having to check through all categories.

nmanduj1:
@seerocode yup, we can link up tomorrow to figure out what that table should look like. I don't think that blocks this pr from being merged tho. It looks good to me!

nmanduj1 (approved):

nmanduj1:
@seerocode When you get a chance can u review my code on this branch ? thank you!!!! (even if not an official review bc its your branch... )

api/schema.graphql
```diff
@@ -139,10 +138,16 @@ input CreateRecordInput {
 }
 
 input EntryInput {
+  id: ID
+  category: CategoryInput!
```
seerocode: I think this can go back to this commit: https://github.com/stanford-policylab/bbc-50-50/pull/62/commits/965bf03d20ba2c43f8fc2213895b368042739ee4 because we are only inputting a category id here as a reference to the category the entry is inputting data for. @nmanduj1

api/mutations.py
```diff
@@ -123,10 +127,15 @@ def resolve_update_record(obj, info, input):
         existing_entry = session.query(Entry).get(entry.get('id'))
 
         if existing_entry:
-            n_entry = Entry(**entry)
-            session.merge(n_entry)
+            category = entry.pop('category')
+            n_category = Category(**category)
+            session.merge(n_category)
+            session.merge(Entry(category=n_category, **entry))
```
seerocode: @nmanduj1 similar here, we don't need to create a category object for the entry as the category id should be sufficient to pull up the relationship later in a record to a Category due to this relationship created here: https://github.com/stanford-policylab/bbc-50-50/pull/62/files#diff-6a0d8c37c66441e7248ec676d7eb0ee364b9f747350c1d71e3f4fdcf448c029eR248. The only thing we are changing as part of an entry creation or update is that instead of a category value and category, we are inputing a category id.

api/schema.graphql
nmanduj1: yup, i modified this thinking we'd need to create/update an Entry's category values when a Record was updated (bc that was the existing functionality when an Entry was updated) But with the abstraction we discussed, that is no longer the case. 

These modifications essentially replicate our past behavior when updating/creating Records and subsequently updating/creating Entries. I am thinking of updating the logic for the new way of handling Categories and Category CRUD on a separate branch. Are you ok merging this branch in as is and having a new branch make these updates? @seerocode

api/database.py
```diff
@@ -335,12 +341,19 @@ def create_dummy_data(session):
...
     ds1.records.append(record)
 
-    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', category_id='51349e29-290e-4398-a401-5bf7d04af75e', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    entry2 = Entry(id='a37a5fe2-1493-4cb9-bcd0-a87688ffa409', category_id='0034d015-0652-497d-ab4a-d42b0bdf08cb', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    entry3 = Entry(id='423dc42f-4628-40e4-b9cd-4e6e9e384d61', category_id='d237a422-5858-459c-bd01-a0abdc077e5b', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    entry4 = Entry(id='407f24d0-c5eb-4297-9495-90e325a00a1d', category_id='662557e5-aca8-4cec-ad72-119ad9cda81b', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    entry5 = Entry(id='4adcb9f9-c1eb-41ba-b9aa-ed0947311a24', category_id='1525cce8-7db3-4e73-b5b0-d2bd14777534', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
-    entry6 = Entry(id='1c49c64f-51e6-48fe-af10-69aaeeddc55f', category_id='a72ced2b-b1a6-4d3d-b003-e35e980960df', count=1, record_id='742b5971-eeb6-4f7a-8275-6111f2342bb4') 
+    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', count=1)
```
seerocode, line 344: Awesome! I see back populating makes creating dummy data a lot easier @nmanduj1 .

api/mutations.py
```diff
@@ -149,4 +149,59 @@ def resolve_delete_record(obj, info, id):
...
+@mutation.field("createCategory")
+@convert_kwargs_to_snake_case
+def resolve_create_category(obj, info, input):
+    '''GraphQL mutation to create a Category.
+        :param input: params for new Category
+        :returns: Category dictionary
+    '''
+
+    session = info.context['dbsession']
+    existing_category = session.query(Category).filter(Category.category_value == input["category_value"].capitalize().strip()).first()
```
seerocode, line 163: I wonder if it would be easier to do this capitalization in the resolver instead of the db class so you don't have to transform the input category value before querying for it? @nmanduj1

seerocode:
@nmanduj1 everything's working as expected! 🥳

api/mutations.py
seerocode: I'm going to merge this since it is all working and checks out!

PR #63, opened 2021-05-24
[ch229] FE - Populate data entry form dynamically using advanced dataset metadata

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/229

image

### **Changes:**
- Updated mock resolvers and tests for metadata addition 
- Updated form render function to render form on add and edit given metadata
- Update client schema for changes (generates auto updated files)

### DIFF
--- a/client/src/__mocks__/MockResolvers.tsx
+++ b/client/src/__mocks__/MockResolvers.tsx
@@ -134,7 +134,7 @@ export const mockResolvers = {
     name: "BBC News",
     datasets: Array.from(datasets.values()),
   }),
-  Date: () => {
+  DateTime: () => {
     return "12/20/2020";
   },
 };
--- a/client/src/__mocks__/MockResolvers.tsx
+++ b/client/src/__mocks__/MockResolvers.tsx
@@ -74,7 +74,11 @@ const categoryData = [
 const recordsData = [
   {
     id: "05caae8d-bb1a-416e-9dda-bb251fe474ff",
-    publicationDate: "2020-12-20",
+    publicationDate: "2020-12-20T00:00:00",
+    dataset: {
+      id: "5a8ee1d5-2b5a-49db-b466-68fe50a27cdb",
+      name: "Breakfast Hour",
+    },
     entries: categoryData,
   },
 ];
@@ -106,10 +110,14 @@ export const mockResolvers = {
   Query: () => ({
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    dataset: (parent, args, ctx, info) => datasets.get(args.id) || {},
+    dataset: (parent, args, ctx, info) => datasets.get(args.id),
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
     user: () => users.get("1"),
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    record: (parent, args, ctx, info) =>
+      recordsData.find((item) => item.id === args.id),
   }),
   Mutation: () => ({
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
--- a/client/src/__mocks__/MockResolvers.tsx
+++ b/client/src/__mocks__/MockResolvers.tsx
@@ -15,56 +15,56 @@ const users = new Map([
 // Mocked categories and values
 const categoryData = [
   {
-    id: "1",
+    id: "a7122153-6cb7-4ea0-a9bd-05dab5a9b293",
     category: {
-      id: "1",
-      category: "Gender",
-      categoryValue: "Cisgender Men",
-    },
-    count: 0,
-  },
-  {
-    id: "2",
-    category: {
-      id: "2",
-      category: "Gender",
-      categoryValue: "Cisgender women",
-    },
-    count: 0,
-  },
-  {
-    id: "3",
-    category: {
-      id: "3",
+      id: "51349e29-290e-4398-a401-5bf7d04af75e",
       category: "Gender",
       categoryValue: "Non-binary",
     },
     count: 0,
   },
   {
-    id: "4",
+    id: "6196eda7-344c-42f9-a367-59cab0b8de33",
     category: {
-      id: "4",
+      id: "0034d015-0652-497d-ab4a-d42b0bdf08cb",
       category: "Gender",
-      categoryValue: "Gender non-conforming",
+      categoryValue: "Cisgender women",
     },
     count: 0,
   },
   {
-    id: "5",
+    id: "5dfb2847-ef6a-44f9-bd54-547b5e018e90",
     category: {
-      id: "5",
+      id: "d237a422-5858-459c-bd01-a0abdc077e5b",
+      category: "Gender",
+      categoryValue: "Cisgender men",
+    },
+    count: 0,
+  },
+  {
+    id: "b1f9b316-ed18-4e80-bc90-2a4702c68cd5",
+    category: {
+      id: "662557e5-aca8-4cec-ad72-119ad9cda81b",
       category: "Gender",
       categoryValue: "Trans women",
     },
     count: 0,
   },
   {
-    id: "6",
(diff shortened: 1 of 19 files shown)


client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -49,13 +50,15 @@ const renderForm = (
       categoryId: record.category.id,
       category: record.category.category,
       categoryValue: record.category.categoryValue,
+      categoryValueLabel: record.category.categoryValue.replace(/\s+/g, "-"),
```
seerocode, line 52: Aria labelled-by and id treats spaces as separate labels so we would need to hyphenate so screen readers can connect the label to the appropriate input.

client/src/__mocks__/MockResolvers.tsx
```diff
@@ -17,27 +16,69 @@ const users = new Map([
...
-  { id: "5", category: "gender", categoryValue: "cisgender", count: 0 },
-  { id: "6", category: "gender", categoryValue: "transgender", count: 0 },
 ];
 
 // Mocked records
 const recordsData = [
   {
     id: "05caae8d-bb1a-416e-9dda-bb251fe474ff",
-    publicationDate: "2020-12-20",
+    publicationDate: "2020-12-20T00:00:00",
```
jnu, line 77: (made this comment on the commit on accident): 

i forget, does the real api return a timezone with the date or no?

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -29,50 +31,39 @@ interface FormProps {
...
+      categoryValue: record.category.categoryValue,
+      categoryValueLabel: record.category.categoryValue.replace(/\s+/g, "-"),
+      count: record.count,
+    }));
+  } else {
+    form = metadata?.dataset.program.targets.map((target) => ({
+      categoryId: target.category.id,
+      category: target.category.category,
+      categoryValue: target.category.categoryValue,
+      categoryValueLabel: target.category.categoryValue.replace(/\s+/g, "-"),
```
jnu, line 59: could use a slightly simpler syntax:
```ts
metadata.dataset.program.targets.map(({category}) => ({
  ...category,
  categoryValueLabel: category.categoryValueLabel.replace(/\s+/g, "-"),
  count: 0,
}));
```

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -163,7 +90,59 @@ const DatasetDetailsRecordsTable = ({
...
       loading={isLoading}
       rowKey={(record) => record.id}
-    />
+    >
+      <Table.Column<TableData>
+        title="Date"
+        dataIndex="publicationDate"
+        key="id"
+        defaultSortOrder="descend"
+        sorter={(dateA: any, dateB: any) =>
```
jnu: do we have a more specific type we can use than `any`?

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -163,7 +90,59 @@ const DatasetDetailsRecordsTable = ({
...
+          dayjs(dateA.publicationDate).unix() -
+          dayjs(dateB.publicationDate).unix()
+        }
+        fixed={true}
+        width={120}
+        render={(date: string) => dayjs(date).format("YYYY-MM-DD")}
+      />
+      {datasetData?.dataset.program.targets.map((target) => (
+        <Table.Column<TableData>
```
jnu, line 108: this syntax blew my mind at first! haven't seen a generic inside JSX before i guess! very cool. :p

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -163,7 +90,59 @@ const DatasetDetailsRecordsTable = ({
...
+          title={target.category.categoryValue}
+          dataIndex={target.category.categoryValue}
+          key="id"
```
jnu, line 111: this key needs to be unique across all the Table.Column elements in this list - maybe you can use `target.id`? Otherwise just using the index from the map function is fine.

client/src/__tests__/Home.test.tsx
```diff
@@ -38,7 +38,7 @@ async function wait(ms = 0) {
 
 test("should render home page datasets and formatted 'last updated' date", async () => {
   const mock = {
-    Date: () => {
+    DateTime: () => {
       return "2021-05-10T03:04:59";
```
jnu, line 42: should this have a timezone associated with it?

seerocode:
@jnu no, we only return the date and time: https://github.com/stanford-policylab/bbc-50-50/blob/main/api/app.py#L82.

We also have DateTime set in the DB but haven't turned on timezone.

client/src/__mocks__/MockResolvers.tsx
```diff
@@ -17,27 +16,69 @@ const users = new Map([
...
-    categoryValue: "gender non-conforming",
+    category: {
+      id: "4",
+      category: "Gender",
+      categoryValue: "Gender non-conforming",
+    },
+    count: 0,
+  },
+  {
+    id: "5",
```
jnu: smol thing - all these objects use uuids now, so switching these too would make things align better!

jnu (approved):
this looks really great to me! nice job, this is exciting!!

client/src/__mocks__/MockResolvers.tsx
seerocode: copy/pasta! no, we only return the date and time: https://github.com/stanford-policylab/bbc-50-50/blob/main/api/app.py#L82.

We also have DateTime set in the DB but haven't turned on timezone.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
seerocode: Thank you! Spread operator always a plus.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: 😆 we actually don't have to do it this way. We could render the columns in an array dynamically like before but essentially this is the same thing as: 
```
const columns: ColumnsType<TableData> = [..]
```

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: String "id" here gets compiled as the unique id of the object here: https://github.com/stanford-policylab/bbc-50-50/pull/63/files/62e72e2471d6ae33e95dde81cda023fee60ca3e2#diff-a460e8deaf5400eaeaca5008fd05ed553748dfe3e1d34d03e5384b6eb9a5551fL45

This is how we're able to pass the ID of an existing record to an entry form on edit or delete it from the table.

The ant design demo for the JSX style API helped me understand: https://ant.design/components/table/#components-table-demo-jsx

client/src/__tests__/Home.test.tsx
seerocode: We don't return a timezone currently from the API but we seems like we'll need to given the time difference. 😆

client/src/__mocks__/MockResolvers.tsx
seerocode: Will update.

client/src/components/DatasetDetails/DatasetDetailsRecordsTable.tsx
seerocode: This is actually `TableData` so I can change that.

client/src/__mocks__/MockResolvers.tsx
seerocode: Tagging @nmanduj1 in case we want to discuss this week.

PR #64, opened 2021-05-27
[ch228] "Last Updated" date appears as "Invalid Date" if no data has been submitted for a dataset

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/228

### Changes:
- Added test to catch when a dataset has no update date
- Added translation text for "no data available"

image

### DIFF
--- a/client/src/__tests__/Home.test.tsx
+++ b/client/src/__tests__/Home.test.tsx
@@ -26,6 +26,16 @@ i18next.use(initReactI18next).init({
   react: {
     useSuspense: false,
   },
+  resources: {
+    "en-gb": {
+      translation: {
+        noDataAvailable: "No Data Available",
+      },
+    },
+    en: {
+      noDataAvailable: "No Data Available",
+    },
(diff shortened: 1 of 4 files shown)


jnu (approved):
:D looks great!

client/src/__tests__/Home.test.tsx
```diff
@@ -26,6 +26,16 @@ i18next.use(initReactI18next).init({
   react: {
     useSuspense: false,
   },
+  resources: {
+    "en-gb": {
```
nmanduj1, line 30: clarification q: "en-gb" == british english?  @seerocode

seerocode: Yeah! This isn't necessary here because we aren't testing for the UK translation specifically in this test. It would be a good to add this to the test setup and create a helper to pass in any resources a test may need so we don't need to copy/paste this whole thing every time.

PR #65, opened 2021-06-02
Refactor i18n tests

seerocode:
Small change to avoid duplicating code for i18next instance in every test.


jnu (approved):
sorry i missed this, thanks for doing!

PR #67, opened 2021-06-04
Ch98/userauth

5harad:
Added backend support for user authentication.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -52,7 +52,21 @@ pip install -r requirements.txt
 uvicorn app:app --reload
 ```
 
+3. Usage
+
 Navigate to http://localhost:8000/graphql/ to see your GraphQL Playground in action!
+
+**Important** Most queries / mutations in the GraphQL schema require permissions to run.
+In development mode, you can use the `X-User` header to supply a user's email address who
+you'd like to run the query as. For example, in the GraphQL playground you can add the headers:
+```
+{
+    "X-User": "tester@notrealemail.info"
+}
+```
+to execute queries as one of the test users.
+See the dummy data in `database.py` for more users, or create your own.
+
 Type in your queries/mutations such as:
 
 ```graphql
--- a/api/app.py
+++ b/api/app.py
@@ -6,7 +6,7 @@ import datetime
 from fastapi import FastAPI, Request
 # from fastapi_sqlalchemy import DBSessionMiddleware  # middleware helper
 # from fastapi_sqlalchemy import db  # an object to provide global access to a database session
-from fastapi_users.authentication import JWTAuthentication
+from fastapi_users.authentication import CookieAuthentication
 from fastapi_users import FastAPIUsers
 from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base
 import dateutil.parser
@@ -14,21 +14,23 @@ import dateutil.parser
 from ariadne import load_schema_from_path, make_executable_schema, snake_case_fallback_resolvers, ObjectType, ScalarType
 from ariadne.asgi import GraphQL
 
-from database import database, SessionLocal
+from database import database, SessionLocal, User
 from queries import queries
 from mutations import mutation
 from settings import settings
 import user
+import directives
 
 
-app = FastAPI()
+app = FastAPI(debug=settings.debug)
 
+# remove cookie_secure=False later for production
+cookie_authentication = CookieAuthentication(secret=settings.secret, lifetime_seconds=3600, cookie_secure=False)
 
-jwt_authentication = JWTAuthentication(secret=settings.secret, lifetime_seconds=3600, tokenUrl="/auth/jwt/login")
 
 fastapi_users = FastAPIUsers(
     user.user_db,
-    [jwt_authentication],
+    [cookie_authentication],
     user.User,
     user.UserCreate,
     user.UserUpdate,
@@ -51,7 +53,7 @@ def after_verification_request(user: user.UserDB, token: str, request: Request):
 
 # Add restful routers for user management
 app.include_router(
-    fastapi_users.get_auth_router(jwt_authentication), prefix="/auth/jwt", tags=["auth"]
+    fastapi_users.get_auth_router(cookie_authentication), prefix="/auth/cookie", tags=["auth"]
 )
 app.include_router(
     fastapi_users.get_register_router(on_after_register), prefix="/auth", tags=["auth"]
@@ -89,7 +91,12 @@ def parse_datetime(value: str) -> datetime.datetime:
 # Adds graphql schema and mounts schema + resolvers to fastapi app
 type_defs = load_schema_from_path("schema.graphql")
 schema = make_executable_schema(
-    type_defs, queries, mutation, datetime_scalar, snake_case_fallback_resolvers
+    type_defs, 
+    queries, 
+    mutation, 
+    datetime_scalar, 
+    snake_case_fallback_resolvers,
+    directives={"needsPermission": directives.NeedsPermissionDirective}
 )
 
 @app.middleware("http")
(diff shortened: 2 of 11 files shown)


api/README.md
```diff
@@ -52,7 +52,21 @@ pip install -r requirements.txt
...
 Navigate to http://localhost:8000/graphql/ to see your GraphQL Playground in action!
+
+**Important** Most queries / mutations in the GraphQL schema require permissions to run.
+In development mode, you can use the `X-User` header to supply a user's email address who
+you'd like to run the query as. For example, in the GraphQL playground you can add the headers:
+```
+{
+    "X-User": "tester@notrealemail.info"
+}
+```
```
seerocode, line 66: 🙏🏽 I would have definitely been confused without this. Thank you!

api/app.py
```diff
@@ -103,14 +110,24 @@ async def add_db_session(request: Request, call_next):
...
+    if settings.debug and "X-User" in request.headers:
+        user = dbsession.query(User).filter(User.email == request.headers["X-User"]).first()
+    else:
+        user = await fastapi_users.current_user(active=True, optional=True)()
+        
     return {
-            "dbsession": request.scope['dbsession'],
+            "dbsession": dbsession,
             "request": request,
+            "current_user": user
```
seerocode, line 126: How do we use this cookie or would we use this cookie on the frontend?

api/database.py
```diff
@@ -75,7 +88,7 @@ class Organization(Base):
     deleted = Column(TIMESTAMP)
 
 
-class Team(Base):
```
seerocode, line 78: Do all the classes need the `PermissionsMixin` or only the ones that were changed in this PR?

api/database.py
```diff
@@ -245,7 +282,15 @@ class Record(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
-class Entry(Base):
+    @classmethod
+    def get_not_deleted(self, session, id_):
+        return session.query(Record).filter(Record.id == id_, Record.deleted == None).first()
+
+    def user_is_team_member(self, user):
```
seerocode, line 336: Neat! 💯 Is this to determine if a user belongs to a queried team?

api/schema.graphql
```diff
@@ -1,25 +1,36 @@
+directive @needsPermission(
+  permission: [Permission!]
+) on FIELD_DEFINITION | OBJECT
+
+enum Permission {
```
seerocode, line 15: Sweet! I wish we could re-order files in PRs because I was looking for these enums in the `directive` file.

api/app.py
jnu: L121 calls into the `fastapi-users` library which fetches the current user based on the cookie. When we write the login UI, the response from a successful login request will set the appropriate cookie, and all the GraphQL requests should just work like normal -- we shouldn't need to modify the apollo client on the FE at all, beyond maybe setting a `credentials: same-origin` option if we haven't set that already.

api/app.py
```diff
@@ -3,32 +3,34 @@
 import sqlalchemy
 import datetime
 
-from fastapi import FastAPI, Request
+from fastapi import FastAPI, Request, Depends
```
nmanduj1: Are you using Depends anywhere?

api/database.py
jnu: If we modify the schema later to use the `TEAM_MEMBER` permission elsewhere, or want to expand the permissions to encompass more things, then maybe we'll need to add the PermissionsMixin to other classes.

If there's a problem somehow where `TEAM_MEMBER` is getting checked on a class that doesn't have this mixin, I set up the directive to throw an error (which is effectively the same as this default `False` permission, just more informative).

api/app.py
```diff
@@ -3,32 +3,34 @@
...
+from database import database, SessionLocal, User
 from queries import queries
 from mutations import mutation
 from settings import settings
 import user
+import directives
 
 
-app = FastAPI()
+app = FastAPI(debug=settings.debug)
```
nmanduj1, line 25: I'm curious to learn more about how fastapi uses this `debug` argument - i can't seem to find anything about it in the fastapi docs, can you link me please?

api/database.py
jnu: yep! the alternative would be to write something like:

```py
if isinstance(obj, Entry):
  # do the check for Entry
elif isinstance(obj, Record):
  # do the check for Record
elif ...
```
in the directive code. It's better OOP style to encapsulate the logic of "how to test this permission for this given class" in each of the respective classes.

api/app.py
jnu: oop, nice catch! removed, thanks

api/app.py
jnu: most web server frameworks use the debug option mostly just to control the amount of info they log. i'm not sure if FastAPI does anything special with the debug option (i also don't see it in their docs), but Starlette (which FastAPI is built on top of) uses it to control how it prints exception stack traces: https://www.starlette.io/applications/

api/database.py
```diff
@@ -124,13 +142,15 @@ class Role(Base):
...
 
+
 class Program(Base):
     __tablename__ = 'program'
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
     name = Column(String(255), nullable=False)
     description = Column(String(255), nullable=False)
     team_id = Column(GUID, ForeignKey('team.id'), index=True)
+    team = relationship('Team', back_populates='programs')
```
nmanduj1: Caught this in my latest pr too! hah

api/database.py
```diff
@@ -227,7 +256,15 @@ class Dataset(Base):
                      server_default=func.now(), onupdate=func.now())
     deleted = Column(TIMESTAMP)
 
-class Record(Base):
+    @classmethod
+    def get_not_deleted(self, session, id_):
```
nmanduj1: Are you moving this from the resolver to the class to avoid repetition?

jnu: yep, exactly! since the logic is a little nuanced and we're repeating it multiple places it's better to put it in a method.

nmanduj1: If so, instead of implementing this method, should we use the concept of `default_scope` (in this case, to fetch non-deleted records) so we can just chain other query clauses if needed?

api/directives.py
```diff
@@ -0,0 +1,198 @@
...
+        GraphQLObjectType,
+        )
+from database import (
+        PermissionsMixin,
+        User,
+        Dataset,
+        Record,
+        )
+
+from typing import List, Iterable
```
nmanduj1, line 14: What's typing do?

api/schema.graphql
```diff
@@ -1,25 +1,36 @@
...
+type Team @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   id: ID!
   name: String!
   users: [User!]!
   programs: [Program!]!
   organizations: [Organization!]!
 }
 
-type User {
+type User @needsPermission(permission: [ADMIN, CURRENT_USER]) {
```
nmanduj1, line 52: What's the difference between an `ADMIN`, a `TEAM_MEMBER` and a `CURRENT_USER`? I guess specifically, an `ADMIN` can be a `CURRENT_USER` non? and how does `LOGGED_IN` fit into this? Can I snag you to run me through these at some point, i think I'm just not following!

jnu: I added documentation here in the schema. You're right that the definitions overlap in some cases -- the permissions check will return true if _any_ of the listed permissions match (not _all_ of them). our primary use case is letting an admin _or_ a user on a certain team do something, and this lets us solve that problem easily.

`LOGGED_IN` just means a user is logged in as _anyone_, which can protect future routes related to reporting that let any user view certain charts as long as they're logged in. It's not really doing much work in the current schema, just enforcing that most of our queries can't be run by a user who isn't logged in (which adds another layer of protection / damage control if there's a bug somewhere in our permissions checks).

jnu: i'm happy to talk this over over zoom if you want!

api/database.py
jnu: i'm not familiar with the `default_scope` - could you point me to docs?

api/directives.py
jnu: this is the python standard library for adding static types to code. `List[Foo]` means a list of type foo, `Iterable[Foo]` means anything that implements the `__iter__` interface, like a list or a set.

we haven't been strict about checking types in python, i feel like it's nice to have if we have time but i don't want to get bogged down in it at this point.

seerocode (approved):
🚢

PR #68, opened 2021-06-07
Ch247/implement description

nmanduj1:


api/queries.py
```diff
@@ -69,4 +70,17 @@ def resolve_category(obj, info, id):
...
+@query.field("value")
+def resolve_value(obj, info, id):
+    '''GraphQL query to find a Value based on Value ID.
+        :param id: Id for the Value to be fetched 
+        :returns: Value dictionary OR None if Value was deleted
+    '''
+    session = info.context['dbsession']
+    value = session.query(Value).get(id)
+    # if not description:
+    #     raise NoResultFound(f'Description with id: {id} was not found')
```
seerocode: Is this a straggler comment?

api/schema.graphql
```diff
@@ -64,14 +64,13 @@ type Target {
   id: ID!
   # Assumption: all datasets in a program have the same targets
   program: Program!
-  category: Category!
+  value: Value!
```
seerocode: What does value mean again? Is this like 'non-binary' for category?

api/database.py
```diff
@@ -185,27 +185,43 @@ class Category(Base):
     __tablename__ = 'category'
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
-    description = Column(String(255), nullable=False)
-    category = Column(String(255), nullable=False)
-    category_value = Column(String(255), nullable=False)
+    name = Column(String(255), nullable=False)
+    description = Column(Text, nullable=False)
+
+    value = relationship('Value', back_populates='category')
```
jnu: this will end up being a list of values, right? maybe the plural `values` would be a better name?

api/database.py
```diff
@@ -172,8 +172,8 @@ class Target(Base):
     program_id = Column(GUID, ForeignKey('program.id'), index=True)
     target_date = Column(DateTime, nullable=False)
     target = Column(Float, nullable=False)
-    category_id = Column(GUID, ForeignKey('category.id'), index=True)
-    category = relationship('Category', back_populates='targets')
+    value_id = Column(GUID, ForeignKey('value.id'), index=True)
+    value = relationship('Value', back_populates='targets')
```
jnu: i wonder if this would be more intuitive as `category_value` or something? i worry that the name `value` might not seem logical in the context of `Target` -- my first impression of the "value" of a target might be the numerical goal instead of the category it's set for.

api/database.py
```diff
@@ -250,8 +266,8 @@ class Entry(Base):
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
 
-    category_id = Column(GUID, ForeignKey('category.id'), index=True)
-    category = relationship('Category', back_populates='entries')
+    value_id = Column(GUID, ForeignKey('value.id'), index=True)
+    value = relationship('Value', back_populates='entries')
```
jnu: Same comment as for `Target.value` -- should we generally call this the `CategoryValue` and `category_value` instead of simply `Value`? maybe it's just me who's confused by this :)

api/database.py
```diff
@@ -303,39 +319,51 @@ def create_dummy_data(session):
...
+    value_non_binary.category = category_gender
+    value_cis_women.category = category_gender
+    value_cis_men.category = category_gender
+    value_trans_women.category = category_gender
+    value_trans_men.category = category_gender
+    value_non_conforming.category = category_gender
+    
+    value_white.category = category_race
+    
+    session.add(category_gender)
```
jnu: not a big deal, but i thought that as long as the relationships between all these objects were defined, you didn't have to explicitly `add` them all to the session?

there's no problem here, my only concern is that this dummy data is starting to get kind of sprawling. i'm happy to brainstorm about how to clean it up, but not a big priority!

api/mutations.py
```diff
@@ -101,8 +102,8 @@ def resolve_create_record(obj, info, input):
     n_entries = []
 
     for entry in all_entries:  
-        category = session.query(Category).get(entry['category_id'])
-        n_entries.append(Entry(category=category, **entry))
+        value = session.query(Value).get(entry['value_id'])
+        n_entries.append(Entry(value=value, **entry))
```
jnu: i believe you can save yourself the query on 105 by just setting the `value_id` here, is that right? it's been a while since i've done this in sql alchemy, but under the hood the database is only storing the `value_id` in the `entry` table anyway, so we shouldn't need to fetch the whole `Value` from the database.

api/mutations.py
```diff
@@ -199,9 +202,59 @@ def resolve_delete_category(obj, info, id):
     session = info.context['dbsession']
 
     session.query(Category).filter(Category.id == id).update({'deleted':datetime.datetime.now()})
-    session.query(Entry).filter(Entry.category_id == id).update({'deleted':datetime.datetime.now()})
-    session.query(Target).filter(Target.category_id == id).update({'deleted':datetime.datetime.now()})
+    session.query(Value).filter(Value.category_id == id).update({'deleted':datetime.datetime.now()})
```
jnu: just want to note that the TIMESTAMP type in sql can behave a little unpredictably when you pass in times. it's a bit better not to calculate the time in python and let the database itself compute it -- you can do this by passing in:

```py
from sqlalchemy.sql import func

session.query(Value).filter(Value.category_id == id).update({'deleted': func.now()})
```

the reason is that the DB and the Python code might be running in different timezones, and the DB itself doesn't store any timezone information with the TIMESTAMP type. so when you read that data back into Python you have to make a judgment call about what timezone it's referring to. it's best if we can assume everything in the db is in the db's timezone -- so here we should just tell the db to figure out what time it is.

api/queries.py
```diff
@@ -69,4 +70,17 @@ def resolve_category(obj, info, id):
     session = info.context['dbsession']
     category = session.query(Category).filter(Category.id == id, Category.deleted == None).first()
     
-    return category
\ No newline at end of file
+    return category
+
+@query.field("value")
```
jnu: [repeating this comment again, sorry!] Query.value seems like it'll be confusing, but i'll defer to you all!

jnu (commented):
made a few comments! i'm happy to explain any of them in detail and/or pair on any other things you have in flight here. overall it looks very thorough and solid, great work!

api/mutations.py
nmanduj1: Great catch! I just learned something new!! I will revisit this where necessary- I'm thinking some of these can be removed entirely given the changes we've made. Thanks joe!!

api/database.py
nmanduj1: I will tinker with it- I am sure there are ways to clean this up. It just wasn't a priority, but you're right that we should keep our code tight!

api/queries.py
nmanduj1: For sure! I was also debating- naming can be so hard!!  I updated this to `category_value`

api/mutations.py
nmanduj1: Thanks for this explanation, Joe! I will update this- time has always been hard for me to wrap my head around. I should spend some time reading into it

api/schema.graphql
nmanduj1: I updated this to `category_value`, like we discussed yesterday 👍

api/queries.py
nmanduj1: Yea, Its from my figuring out how to elegantly handle error messaging. sorry! I have a few comments laying around- I will clean them up before opening a formal PR. Thank you for checking out my draft!!

api/mutations.py
nmanduj1: Hmm, @jnu I updated this and it seems I also have to add a `synchronize_session ` flag to `.update()` in order for that `.update()` to run as expected. My inclination is to pass a `False` flag, given that we are (currently) always using a `.commit()` immediately after updating. In the event we are continuing to use the objects in-memory, we'd probably want that flag to be set to `fetch`-

api/mutations.py
```diff
@@ -160,8 +143,7 @@ def resolve_create_category(obj, info, input):
     '''
 
     session = info.context['dbsession']
-    existing_category = session.query(Category).filter(Category.category_value == input["category_value"].capitalize().strip()).first()
-
+    existing_category = session.query(Category).filter(Category.name == input["name"].capitalize().strip()).first()
```
jnu: i have three suggestions here:
1) the name cleaning logic is duplicated here and in the `Category` field validator you wrote (which btw is awesome! validators are great!). This means we have to keep the logic in sync in both places, so if we ever found a bug or wanted to change the implementation we'd have to remember to update both functions. It's better to keep this all in one single function so that it's simpler to maintain if we want to change the implementation. And speaking of changing the implementation ... :p
2) a subtle thing, the order of the string manipulation matters here:
```py
name.capitalize().strip() != name.strip().capitalize()
```
we want to strip _before_ we capitalize, otherwise if there's whitespace at the beginning of the name the capitalization will have no effect
3) lastly, as you saw in my PR, I'm a fan of adding helper methods to the database.py classes to encapsulate common query patterns. In this way you can consolidate all the logic about how to standardize / clean / validate names in the `Category` class itself. This is good object-oriented practice and simplifies maintaining the code in the long run.

I would refactor to do something like:
```py
class Category(Base):
    @classmethod
    def get_by_name(cls, session, name):
        return session.query(cls).filter(cls.name == cls.clean_name(name)).first()

    @classmethod
    def clean_name(cls, name):
        return name.capitalize().strip()

    @validates('name')
    def capitalize_category(self, key, name):
        # NOTE: `self.__class__` basically just means `Category` here. It's slightly
        # better to avoid referencing the class explicitly by name in case a) we change
        # the name of the class, or b) we extend the class and want to allow the child
        # to override the method.
        return self.__class__.clean_name(name)
```

then here you can call:

```py
existing_category = Category.get_by_name(session, input["name"])
```
and now the mutations don't need to know anything about how to standardize the `name` input, it's all handled by the `Category` class itself 😊

api/schema.graphql
```diff
@@ -64,14 +64,13 @@ type Target {
   id: ID!
   # Assumption: all datasets in a program have the same targets
   program: Program!
-  category: Category!
+  category_value: CategoryValue!
```
jnu: smol nit: we should be consistent about case here - generally this schema uses camelCase (e.g. on line 102 in Entry.categoryValue)

api/database.py
```diff
@@ -275,132 +288,68 @@ def create_tables(engine, session):
...
-
-    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', count=1) 
-    entry2 = Entry(id='a37a5fe2-1493-4cb9-bcd0-a87688ffa409', count=1) 
-    entry3 = Entry(id='423dc42f-4628-40e4-b9cd-4e6e9e384d61', count=1) 
-    entry4 = Entry(id='407f24d0-c5eb-4297-9495-90e325a00a1d', count=1) 
-    entry5 = Entry(id='4adcb9f9-c1eb-41ba-b9aa-ed0947311a24', count=1) 
-    entry6 = Entry(id='1c49c64f-51e6-48fe-af10-69aaeeddc55f', count=1) 
-    entry7 = Entry(id='335b3680-13a1-4d8f-a917-01e1e7e1311a', count=1)
-    entry8 = Entry(id='fa5f1f0e-d5ba-4f2d-bdbf-819470a6fa4a', count=1)
+    Record(id='742b5971-eeb6-4f7a-8275-6111f2342bb4', dataset=ds1, publication_date=datetime.strptime('2020-12-21 00:00:00', '%Y-%m-%d %H:%M:%S'),
```
jnu, line 369: wow nice, this is so much fewer lines of code!! love it <3

api/database.py
```diff
@@ -249,12 +261,13 @@ class Entry(Base):
...
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
-
-    category_id = Column(GUID, ForeignKey('category.id'), index=True)
-    category = relationship('Category', back_populates='entries')
+    category_value_id = Column(GUID, ForeignKey('category_value.id'), index=True)
+    category_value = relationship('CategoryValue', back_populates='entries')
     count = Column(Integer, nullable=False)
     record = relationship('Record', back_populates='entries')
     record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
+    inputter = relationship('User')
```
jnu, line 292: just checking, you mentioned we're not filling this column out currently, correct? After we merge #67 we can track inputter with `info.context['current_user']`

jnu (approved):
this looks great nancy, nice work! i made a couple of suggestions that should be quick to address.

Unfortunately this is going to have some painful merge conflicts with #67, especially with the test updates. If you want to merge this this morning I'm happy to solve the conflicts on my end.

PR #69, opened 2021-06-09
Test db connection validity before using

jnu:
Makes sure database connection is still valid before using. more info: https://docs.sqlalchemy.org/en/13/core/pooling.html#disconnect-handling-pessimistic


seerocode (approved):
Thanks for linking to the article that explains this!

nmanduj1 (approved):

PR #70, opened 2021-06-09
[ch165] Summary stats on dataset details page

seerocode:
# Story details: 
https://app.clubhouse.io/stanford-computational-policy-lab/story/165

### Changes:
- added temp pie charts for dataset details stats in score cards
- refactored a few functions/names for clarity 
- updated test to mock canvas chart elements
- all changes have been rebased on top of changes in main branch

### Screenshots:
data added but all zero:
image

number data added:
image

no data:
image

ipad view:
image

iphone x (landscape):
image

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -283,6 +283,11 @@ class CategoryValue(Base):
     @validates('name')
     def capitalize_name(self, key, name):
         return name.capitalize().strip()
+
+    @classmethod
+    def get_not_deleted(cls, session, id):
+        return session.query(CategoryValue).\
+            filter(CategoryValue.id == id, CategoryValue.deleted == None).first()
     
 
 class Dataset(Base, PermissionsMixin):
--- a/api/queries.py
+++ b/api/queries.py
@@ -63,8 +63,8 @@ def resolve_sums_of_category_values(dataset, info):
     '''
     session = info.context['dbsession']
 
-    stmt = session.query(Entry.category_id, func.sum(Entry.count).label('sum_of_counts')).\
-        join(Entry.record).filter(Record.dataset_id == dataset.id, Record.deleted == None).group_by(Entry.category_id).all()
+    stmt = session.query(Entry.category_value_id, func.sum(Entry.count).label('sum_of_counts')).\
+        join(Entry.record).filter(Record.dataset_id == dataset.id, Record.deleted == None).group_by(Entry.category_value_id).all()
 
     return [{'dataset_id': dataset.id, 'category_value_id': row[0], 'sum_of_counts': row[1]} for row in stmt]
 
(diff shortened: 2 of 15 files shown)


nmanduj1 (approved):
woot woot! looks good to me! 🥇

api/database.py
```diff
@@ -283,6 +283,11 @@ class CategoryValue(Base):
     @validates('name')
     def capitalize_name(self, key, name):
         return name.capitalize().strip()
+
+    @classmethod
+    def get_not_deleted(cls, session, id):
+        return session.query(CategoryValue).\
+            filter(CategoryValue.id == id, CategoryValue.deleted == None).first()
```
nmanduj1, line 290: joe used `.scalar() `in one of his prs and I think it makes sense to use `.scalar()` instead of `.first()` throughout- per docs: `.scalar()` "Returns the first element of the first result or None if no rows present. If multiple rows are returned, raises MultipleResultsFound." https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.scalar

i am updating them as i use them in branch ch295.

seerocode: I thought scalar was like the actual object value like it will return "1" instead of (UUID(1)) or something like that?

nmanduj1: hmm, I found this explanation:
```
Query.one() executes the query and raises MultipleResultsFound if the number of rows was more than 1, and NoResultFound no rows were found. Otherwise it returns the sole result row (as tuple)
Query.first() executes the query with LIMIT 1 and returns the result row as tuple, or None if no rows were found
Query.scalar() executes the query and raises MultipleResultsFound if the number of rows was more than 1. If no rows were found, it returns None. Otherwise it returns the first column of the sole result row.
Query.one() should be used if you want to select a single row and you assume there is exactly one row to be found. Query.first() should be used when you want just one row but you're not sure if one will be found. Query.scalar() should be used when you want, for example, just the maximum of a single column.
```
https://alextechrants.blogspot.com/2013/11/10-common-stumbling-blocks-for.html#:~:text=first()%20executes%20the%20query,rows%20was%20more%20than%201.&text=Otherwise%20it%20returns%20the%20first%20column%20of%20the%20sole%20result%20row.

Given this nuance, I think we'd want either a `.one()` or .`scalar()` here (since we'd want an error thrown if multiple results were found- we should not be finding multiple results) However since we don't require a 404 error bc we're using graphql, I think using a .`scalar()` would be our best bet.

nmanduj1: in theory, it's kind of a moot point bc our DB is enforcing uniqueness- there's no way we could insert duplicate db records with the same id. i guess this is more so just coding defensively? i think any of them should work just fine.

seerocode: Ah, okay I remember using scalar to return a single number so that's what I connected. Using `one()` also makes sense to me.

PR #71, opened 2021-06-15
[Ch269] Login UI

jnu:
depends on #67 

Implements a login UI prompt where user can submit email/password to gain access to the site.

In this PR:
 - [x] Login component
 - [x] Refactor routes to redirect to Login component when not authorized
 - [x] Auth service that integrates with API auth routes
 - [x] Logout button in app header
 - [x] Add and update tests

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -371,23 +371,25 @@ def create_tables(engine, session):
 
 
 def create_dummy_data(session):
+    from passlib.hash import bcrypt
+
     print("👩🏽‍💻 Adding dummy data ...")    
     user = User(id='cd7e6d44-4b4d-4d7a-8a67-31efffe53e77',
             email='tester@notrealemail.info',
-            hashed_password='$2b$12$jsOYCT1t0H6T3fJ/ODmySO46/Z1zHqYauAbFsgpyjUG1ZUWsAQwn6',
+            hashed_password=bcrypt('password'),
             first_name='Cat', last_name='Berry')
     
     # Secondary app user (no perms, used for testing access controls)
     other_user = User(id='a47085ba-3d01-46a4-963b-9ffaeda18113',
     email='other@notrealemail.info',
-    hashed_password='c053ecf9ed41df0311b9df13cc6c3b6078d2d3c2',
+    hashed_password=bcrypt('otherpassword'),
     first_name='Penelope', last_name='Pineapple')
     session.add(other_user)
                       
     # Admin user
     admin_user = User(id='df6413b4-b910-4f6e-8f3c-8201c9e65af3',
                     email='admin@notrealemail.info',
-                    hashed_password='c053ecf9ed41df0311b9df13cc6c3b6078d2d3c2',
+                    hashed_password=bcrypt('adminpassword'),
                     first_name='Daisy', last_name='Carrot')
     admin = session.query(Role).get('be5f8cac-ac65-4f75-8052-8d1b5d40dffe')
     admin_user.roles.append(admin)
--- a/api/database.py
+++ b/api/database.py
@@ -374,7 +374,7 @@ def create_dummy_data(session):
     print("👩🏽‍💻 Adding dummy data ...")    
     user = User(id='cd7e6d44-4b4d-4d7a-8a67-31efffe53e77',
             email='tester@notrealemail.info',
-            hashed_password='c053ecf9ed41df0311b9df13cc6c3b6078d2d3c2',
+            hashed_password='$2b$12$jsOYCT1t0H6T3fJ/ODmySO46/Z1zHqYauAbFsgpyjUG1ZUWsAQwn6',
             first_name='Cat', last_name='Berry')
     
     # Secondary app user (no perms, used for testing access controls)
--- a/api/database.py
+++ b/api/database.py
@@ -371,25 +371,32 @@ def create_tables(engine, session):
 
 
 def create_dummy_data(session):
+    if not settings.debug:
+        raise RuntimeError("Can't add dummy data when not in a debug environment")
+
     from passlib.hash import bcrypt
+    def hash_test_password(pw: str) -> str:
+        # NOTE: Use consistent salt and rounds to eliminate non-determinism in
+        # testing. This is obviously not intended to be secure.
+        return bcrypt.hash(pw, salt="0"*22, rounds=4)
 
     print("👩🏽‍💻 Adding dummy data ...")    
     user = User(id='cd7e6d44-4b4d-4d7a-8a67-31efffe53e77',
             email='tester@notrealemail.info',
-            hashed_password=bcrypt('password'),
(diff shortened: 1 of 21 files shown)


client/public/locales/en/translation.json
```diff
@@ -11,5 +11,12 @@
   "addData": "Add Data",
   "editData": "Edit",
   "backToDashboard": "Back To Dashboard",
-  "noDataAvailable": "No Data Available"
+  "noDataAvailable": "No Data Available",
```
seerocode, line 14: These will be the defaults for UK English as well.

client/src/components/Login/Login.tsx
```diff
@@ -0,0 +1,113 @@
...
+ * service that manages authentication.
+ */
+export type LoginProps = { auth: Auth } & RouteComponentProps;
+
+/**
+ * Login UI form.
+ */
+export const Login = (props: LoginProps) => {
+  const { t, i18n } = useTranslation();
+  const [loading, setLoading] = useState(false);
```
seerocode, line 33: Is it okay to leave off the boolean type here or should it be added?

client/src/router/Router.tsx
```diff
@@ -22,11 +24,53 @@ function flattenRoutes(route: IRoute, agg: any = []) {
...
 
+/**
+ * Type representing the base router props.
+ */
+type AnyRouteProps = RouteComponentProps<Record<string, string | undefined>>;
+
+/**
+ * Type representing a React component that can be used as a react-router component.
+ *
+ * The component can be either a classical or a functional component.
```
seerocode, line 36: By classical this means a class component or something else?

client/src/router/Router.tsx
```diff
@@ -37,3 +81,45 @@ export function RenderRoutes({ routes }: any) {
...
+ * authenticated. Once the user is authed, they will have access to the list
+ * of protected routes passed in here.
+ */
+export function RenderRoutes({
+  loginComponent,
+  protectedRoutes,
+  protectedContainer,
+  auth,
+}: RenderRoutesProps) {
+  const C = protectedContainer;
```
seerocode: Does uppercase C mean the protected container? If so, maybe that could be the name instead.

client/src/router/Router.tsx
```diff
@@ -37,3 +81,45 @@ export function RenderRoutes({ routes }: any) {
...
+      <ProtectedRoutes routes={protectedRoutes} />
+    </C>
+  ));
+
+  const L = loginComponent;
+  const WrappedLogin = (props: AnyRouteProps) => <L {...props} auth={auth} />;
+
+  return (
+    <Switch>
+      <Route exact path="/login" component={WrappedLogin} />
```
seerocode: Does this wrap the login and other routes to protect the pages beyond the login page?

client/src/components/Login/Login.tsx
jnu: typescript can infer simple types like this so no need to add explicitly

client/src/router/Router.tsx
jnu: yep! classical = related to classes and object-oriented programming.

client/src/router/Router.tsx
jnu: yea, JSX doesn't let you render custom components that start with a lowercase letter, but props and options are conventionally camelCase (see e.g. Enzyme's `wrappingComponent` in shallow rendering optoins).

I just made the names more clear though

client/src/router/Router.tsx
jnu: Three slightly different reasons for doing the wrapping for two different components:
1) The `protectedRoutes` need to be wrapped with a conditional "if user is not logged in, redirect to login page"
2) The `protectedRoutes` also need some app layout (like our sidebar and header) that should _not_ be shown on the Login screen.

Then there is
3) The `Login` component needs to be wrapped to inject the `auth` service. 

But I just realized that for (3) it's probably simpler now to just use the `useAuth` hook I wrote that can grab the auth object from the `AuthProvider`, so no need for more wrapping! Just updated this, thanks!

jnu:
image


UI is not the prettiest, but I think we should merge soonish and polish later!

api/database.py
```diff
@@ -374,7 +374,7 @@ def create_dummy_data(session):
     print("👩🏽‍💻 Adding dummy data ...")    
     user = User(id='cd7e6d44-4b4d-4d7a-8a67-31efffe53e77',
             email='tester@notrealemail.info',
-            hashed_password='c053ecf9ed41df0311b9df13cc6c3b6078d2d3c2',
+            hashed_password='$2b$12$jsOYCT1t0H6T3fJ/ODmySO46/Z1zHqYauAbFsgpyjUG1ZUWsAQwn6',
```
seerocode: I panicked for a second and saw this was dummy data! LOL

client/src/__tests__/App.test.tsx
```diff
@@ -1,10 +1,30 @@
...
 
 describe("App Component", () => {
-  test("renders without crashing", () => {
-    const wrapper = shallow(<App />);
-    expect(wrapper).toBeTruthy();
+  test("renders correctly", async () => {
+    const { auth, mock } = mockUserLoggedIn();
+    await auth.init();
+
+    const tree = shallow(<App />, {
```
seerocode, line 12: I've moved completely away from enzyme in other tests in favor of react testing library but left it for App cause low priority. I don't think it needs to be changed here right now.

client/src/components/AuthProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    throw auth.init();
+  }
+
+  return <AuthContext.Provider value={auth}>{children}</AuthContext.Provider>;
+};
+
+/**
+ * React hook to get the auth object from context.
+ */
+export const useAuth = () => useContext(AuthContext)!;
```
seerocode, line 23: Awesome that this is a hook!! In case we make other hooks, could this be moved to a hooks directory?

seerocode (approved):
This looks awesome!

api/database.py
jnu: hah--speaking of, i should add a comment explaining what the password is :)

spoiler: it's `password`

client/src/__tests__/App.test.tsx
jnu: ah ok, good to know, thanks!

client/src/components/AuthProvider.tsx
jnu: totally up to you! you've done a great job organizing the frontend code and i'm just trying not to mess it up!

seerocode: We are starting to have more components that go beyond pages so it might be time for a file structure refactor soon so we can separate pages from other components.

PR #72, opened 2021-06-15
Update frontend for category descriptions and user auth

seerocode:
### Changes for categoryvalue update from backend:
- Update mock resolvers for failing tests
- Update dataset records table and entry form render function

### DIFF
--- a/api/app_test.py
+++ b/api/app_test.py
@@ -25,6 +25,8 @@ class BaseAppTest(unittest.TestCase):
     """Base test runner that sets up an in-memory database that test cases can
     make assertions against.
     """
+    # Get full diff outuput
+    maxDiff = None
 
     def setUp(self):
         # Create in-memory sqlite database. This is configured to work with
@@ -870,7 +872,6 @@ class TestGraphQL(BaseAppTest):
                 },
             }, user=user)
 
-            # self.maxDiff = None
             self.assertTrue(success)
             self.assertTrue(self.is_valid_uuid(result["data"]["createRecord"]["id"]), "Invalid UUID")
             self.assertEqual(result, {
--- a/api/app_test.py
+++ b/api/app_test.py
@@ -886,7 +886,7 @@ class TestGraphQL(BaseAppTest):
                                 "id": "662557e5-aca8-4cec-ad72-119ad9cda81b",
                                 "name": "Trans women"
                             },
-                            "inputter": {"id": result["data"]["createRecord"]["entries"][0]["inputter"]["id"]}
+                            "inputter": {"id": str(user.id)}
                         }],
                     },
                 },
--- a/api/app_test.py
+++ b/api/app_test.py
@@ -865,12 +865,12 @@ class TestGraphQL(BaseAppTest):
                         "entries": [{
                             "count":  7,
                             "categoryValueId": "662557e5-aca8-4cec-ad72-119ad9cda81b",
-                            "inputterId": "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77",
                         }],
                     }
                 },
             }, user=user)
 
+            # self.maxDiff = None
             self.assertTrue(success)
             self.assertTrue(self.is_valid_uuid(result["data"]["createRecord"]["id"]), "Invalid UUID")
             self.assertEqual(result, {
@@ -885,7 +885,7 @@ class TestGraphQL(BaseAppTest):
                                 "id": "662557e5-aca8-4cec-ad72-119ad9cda81b",
                                 "name": "Trans women"
                             },
-                            "inputter": {"id": "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77"}
+                            "inputter": {"id": result["data"]["createRecord"]["entries"][0]["inputter"]["id"]}
                         }],
                     },
                 },
@@ -973,7 +973,6 @@ class TestGraphQL(BaseAppTest):
                         "entries": [{
                             "id": "64677dc1-a1cd-4cd3-965d-6565832d307a",
                             "categoryValueId": "6cae6d26-97e1-4e9c-b1ad-954b4110e83b",
-                            "inputterId": "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77",
                             "count": 0
                             }
                         ]
--- a/api/database.py
+++ b/api/database.py
@@ -213,7 +213,7 @@ class Tag(Base):
     deleted = Column(TIMESTAMP)
 
 
-class Target(Base):
+class Target(Base, PermissionsMixin):
     __tablename__ = 'target'
 
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
--- a/api/mutations.py
+++ b/api/mutations.py
@@ -88,10 +88,12 @@ def resolve_create_record(obj, info, input):
     '''
 
     session = info.context['dbsession']
+    current_user = info.context['current_user']
+
     all_entries = input.pop('entries', [])
     n_entries = []
     for entry in all_entries:  
-        n_entries.append(Entry(**entry))
+        n_entries.append(Entry(inputter=current_user, **entry))
     record = Record(entries=n_entries, **input)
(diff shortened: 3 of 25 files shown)


api/schema.graphql
```diff
@@ -200,7 +200,7 @@ input UpdateCategoryValueInput {
 
 input EntryInput {
   id: ID
-  inputterId: ID!
+  inputterId: ID
```
seerocode: We might not need this here anymore but leaving it be for now.

client/src/index.tsx
```diff
@@ -30,6 +31,17 @@ const httpLink = new HttpLink({
       : "/graphql/",
 });
 
+const authLink = new ApolloLink((operation, forward) => {
+  // Use the setContext method to set the HTTP headers
+  operation.setContext({
+    headers: {
+      "X-User": "tester@notrealemail.info", // TODO: remove placeholder when user login is merged
```
seerocode: Pending merge of login UI work to remove the placeholder.

api/mutations.py
```diff
@@ -88,10 +88,12 @@ def resolve_create_record(obj, info, input):
     '''
 
     session = info.context['dbsession']
+    current_user = info.context['current_user']
```
seerocode, line 91: I moved the inputter id setting to the resolver since there isn't a way to get this info right now from the frontend. Also referencing the comment here: https://github.com/stanford-policylab/bbc-50-50/pull/68#discussion_r651831754

jnu: this makes sense to me! i think this is the right way to go regardless of whether we can get the right ID in the UI or not.

api/schema.graphql
jnu: I think we should actually remove this. currently it's a security problem to leave it in -- we'd want to validate that the `current_user` matches the `inputterId` anyway, which we don't currently do. (otherwise you could enter data as someone else.)

there's a tiny possibility that someone will want permission to enter on behalf of someone else... i don't think we should implement that unless we absolutely have to!

client/src/index.tsx
jnu: glad you got this to work around! if development with a login becomes tedious we can discuss more workarounds to make it easier.

jnu (approved):
nice work!

api/schema.graphql
seerocode: I removed the field in the schema and updated the record tests. CC @nmanduj1

api/app_test.py
```diff
@@ -782,12 +782,12 @@ def test_create_record(self):
                         "entries": [{
                             "count":  7,
                             "categoryValueId": "662557e5-aca8-4cec-ad72-119ad9cda81b",
-                            "inputterId": "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77",
                         }],
                     }
                 },
             }, user=user)
 
+            # self.maxDiff = None
```
seerocode: I'll be removing this in a bit.

api/schema.graphql
```diff
@@ -200,7 +200,6 @@ input UpdateCategoryValueInput {
 
 input EntryInput {
   id: ID
-  inputterId: ID!
```
seerocode, line 203: I removed inputterId from the schema to resolve the possible security issue and we are setting this in the resolver.

client/src/components/Home/Home.tsx
```diff
@@ -102,16 +104,19 @@ const getTableData = (
 
 const Home = (): JSX.Element => {
   const { t } = useTranslation();
+  const userId = useAuth().getUserId();
 
   const { data, loading, error } = useQuery<GetUser, GetUserVariables>(
     GET_USER,
     {
-      variables: { id: "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77" }, // TODO: Replace after user auth implementation
+      variables: { id: userId },
```
seerocode, line 112: This is getting the user id from the currently logged in user.

client/src/index.tsx
```diff
@@ -33,6 +34,17 @@ const httpLink = new HttpLink({
       : "/graphql/",
 });
 
+const authLink = new ApolloLink((operation, forward) => {
+  // Use the setContext method to set the HTTP headers
+  operation.setContext(({ headers = {} }) => ({
+    headers: {
+      "X-User": auth.getEmail(), // get user email from auth service
```
seerocode: @jnu i couldn't get the queries to work without this header. Is this the appropriate way to set this or is there a different method?

api/app_test.py
jnu: it's fine with me if you want to keep the maxDiff setting in place, i'm only _not_ a fan of commented out code lines :) 

you can make it an attribute of the class if you want so it applies to all the test cases:

```py
class TestGraphQL(unittest.TestCase):

    maxDiff = None

    def setUp():
        # etc...
```

api/app_test.py
```diff
@@ -802,7 +802,7 @@ def test_create_record(self):
                                 "id": "662557e5-aca8-4cec-ad72-119ad9cda81b",
                                 "name": "Trans women"
                             },
-                            "inputter": {"id": "cd7e6d44-4b4d-4d7a-8a67-31efffe53e77"}
+                            "inputter": {"id": result["data"]["createRecord"]["entries"][0]["inputter"]["id"]}
```
jnu: can you use `user.id` here? that way it'll assert that the code that infers the inputter automatically is working correctly (as written `inputter['id']` could be anything because we're just comparing it to itself)

api/database.py
```diff
@@ -213,7 +213,7 @@ def capitalize_tag_name(self, key, name):
     deleted = Column(TIMESTAMP)
 
 
-class Target(Base):
+class Target(Base, PermissionsMixin):
```
jnu, line 216: thanks for adding this!

client/src/index.tsx
jnu: hm, queries are working for me without it. this header will be **completely ignored** outside of development mode. so i think we should remove this authLink entirely (since auth is handled by the brother with cookies) ... it's possible it's doing some unintentional meddling like ignoring cookies.

since there's a lot in flight it might be hard to debug what the issue might be. i suggest:
1) remove the authlink, so we don't have a workaround in place that only works in dev but not prod (that'll be harder to debug)
2) merge your PR and my #74 into main
3) we both pull latest main 
4) we both reset our local development databases
5) see if there's still an issue and debug as necessary

how does that sound?

client/src/services/auth.ts
```diff
@@ -68,6 +68,28 @@ export class Auth {
     return `${this.currentUser.first_name} ${this.currentUser.last_name}`;
   }
 
+  /**
+   * The UUID of the currently authenticated user.
+   * @returns {string} UUID
```
jnu, line 82: i'm always on the fence about writing documentation for simple functions in typescript. docs are hard to keep in sync with code, and typescript is already able to tell us that the return type is a string.

jnu (approved):
this looks good! i wrote a couple things that i think we should address before merging:
1) the python test i think needs to use `user.id`
2) the `x-user` header makes me nervous to keep in, since by design it won't work in prod, so we need to tackle that issue head-on now (my suggestion is to remove that authLink entirely and we can see if the issue still happens once we merge everything in to `main`)

great work!!!

client/src/index.tsx
seerocode: OK I removed it and rebased changes from main since #74 was merged in.

api/app_test.py
seerocode: Sure thing! It makes sense to have done this in the first place 🤦🏽

client/src/index.tsx
jnu: > since auth is handled by the brother with cookies

i think that meant to see _browser_ ... 😅

jnu: see = say 

i'm gonna stop typing ...

seerocode: I like brother better lol

jnu: giphy

PR #74, opened 2021-06-16
Check for admin role

jnu:
The built-in fastapi-users `/users` router won't return our custom roles. Didn't see a super obvious way to make fastapi-users work better with the sqlalchemy ORM, so I just added a simple drop-in replacement for the `/users/me` route that returns the `roles`.

This also adds a test on the frontend `Auth#isAdmin` to check whether the user should see the admin portal.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -3,7 +3,7 @@ import databases
 import sqlalchemy
 import datetime
 
-from fastapi import FastAPI, Request
+from fastapi import FastAPI, Request, Depends
 # from fastapi_sqlalchemy import DBSessionMiddleware  # middleware helper
 # from fastapi_sqlalchemy import db  # an object to provide global access to a database session
 from fastapi_users.authentication import CookieAuthentication
@@ -76,7 +76,23 @@ app.include_router(
     prefix="/auth",
     tags=["auth"],
 )
-app.include_router(fastapi_users.get_users_router(), prefix="/users", tags=["users"])
+
+# Use a custom implementation of the /users/me instead of fastapi-users's.
+# Their permissions are too simple for us; we want to return a list of roles
+# for the user. Behavior around 200 and 401 responses is otherwise the same.
+@app.get("/users/me")
+def get_current_user(request: Request, user: user.User = Depends(fastapi_users.current_user(active=True))):
+    session = request.scope["dbsession"]
+    dbuser = session.query(User).get(user.id)
+    return {
+            "id": dbuser.id,
+            "roles": [r.name for r in dbuser.roles],
+            "first_name": dbuser.first_name,
+            "last_name": dbuser.last_name,
+            "email": dbuser.email,
+            "is_active": dbuser.is_active,
+            "is_verified": dbuser.is_verified,
+            }
 
 
 # General Graphql Field Resolvers
--- a/api/app.py
+++ b/api/app.py
@@ -22,7 +22,14 @@ import user
 import directives
 
 
-app = FastAPI(debug=settings.debug)
+app = FastAPI(
+        debug=settings.debug,
+        # NOTE: the following dependencies are available in the `extra` dict
+        # on the app object. They're injected here so they can be patched
+        # easier in testing.
+        db=database,
+        get_db_session=SessionLocal,
+        )
 
 # remove cookie_secure=False later for production
 cookie_authentication = CookieAuthentication(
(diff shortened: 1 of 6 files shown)


api/app.py
```diff
@@ -76,7 +76,23 @@ def after_verification_request(user: user.UserDB, token: str, request: Request):
     prefix="/auth",
     tags=["auth"],
 )
-app.include_router(fastapi_users.get_users_router(), prefix="/users", tags=["users"])
+
+# Use a custom implementation of the /users/me instead of fastapi-users's.
+# Their permissions are too simple for us; we want to return a list of roles
+# for the user. Behavior around 200 and 401 responses is otherwise the same.
+@app.get("/users/me")
+def get_current_user(request: Request, user: user.User = Depends(fastapi_users.current_user(active=True))):
```
seerocode: What does `Depends` mean?

nmanduj1: Yay! Depends!

client/src/services/auth.ts
```diff
@@ -58,6 +60,13 @@ export class Auth {
     return !!this.currentUser;
   }
 
+  /**
+   * Test whether user has admin permissions.
+   */
+  public isAdmin() {
+    return !!this.currentUser && this.currentUser.roles.contains("admin");
```
nmanduj1: q: Will we be differentiating between team-lead admins and lara level admins?

jnu: not now, but possibly in the future, if we need to! we'd just need to add a `Role` and distinguish that here.

api/app.py
jnu: FastAPI uses `Depends` for dependency injection in routes. in other words, it's a way to get a reference to an object inside of the request handler that depends on the context of the request. it's effectively the same as writing a middleware that stores a value in the request context, which is generally how this is done in other frameworks.

api/app_test.py
```diff
@@ -35,8 +45,81 @@ def setUp(self):
...
-        # Create new, fresh session for test case to use
-        self.session = Session()
+        # Expose the session maker so a test can make one
+        self.Session = Session
+
+    def tearDown(self):
+        self.engine.dispose()
+
+
+class TestAppUsers(BaseAppTest):
```
seerocode, line 55: I was thinking this test file is getting pretty long but I guess it doesn't matter right especially if breaking this up is more trouble than it's worth.

api/app.py
```diff
@@ -105,7 +131,7 @@ def parse_datetime(value: str) -> datetime.datetime:
 
 @app.middleware("http")
 async def add_db_session(request: Request, call_next):
-    session = SessionLocal()
+    session = app.extra['get_db_session']()
```
nmanduj1, line 134: What is app.extra doing here?

seerocode (approved):
🚢

api/app_test.py
```diff
@@ -35,8 +45,81 @@ def setUp(self):
...
+    """Test /users/ API routes."""
+
+    def setUp(self):
+        super().setUp()
```
nmanduj1, line 59: What is `.super()` doing here?

nmanduj1: I think it has to be run during set up and teardown- but i dont understand why

api/app.py
jnu: `extra` is a dict of arbitrary values that is created when you create the `app`. the issue this is solving here is that in `database.py` we create the db engine and sessionmaker for _postgres_, but in tests we don't have postgres running. (and even if we did, the connection settings would likely be different than what we use in `database.py` for prod.) instead of referencing the _global_ `database` and `SessionLocal` variables, i changed this to reference whatever values we currently have `app` configured with. this makes writing tests much easier, because we can swap in a mock dependency for the database much more cleanly than we can patch the global objects in `database.py`.

this general principle is called "dependency injection" and it's a very helpful design practice to keep code modular and testable.

there are other (and even nicer) ways of accomplishing this than using the `extra` dict, but i didn't want to do too much refactoring now -- to save time and avoid merge conflicts.

(I made a comment about this above on L27 explaining this too!)

nmanduj1 (approved):
Yay tests !

api/app_test.py
jnu: yep, would be good to organize better in the future, but we should avoid that in this PR to avoid conflicts with yours!

api/app_test.py
jnu: the class that this testcase inherits also defines a `setUp()` method, and we want both to be run.

```py
class Base:
    def hi(self):
        print("hello from base class!")

class Child(Base):
    def hi(self, run_base_hi):
        if run_base_hi:
            super().hi()
        print("hello from child class!")

Child().hi(run_base_hi=False)
# prints "hello from child class!" only
Child().hi(run_base_hi=True)
# prints "hello from base class!" and "hello from child class!"
```

if you don't call `super()` the child class method will override the base class method completely.

PR #75, opened 2021-06-17
Fix user auth

jnu:
Pass in cookie correctly to fastapi_users so current user can be verified. This is a little hacky, would be nice to figure out how to use Depends with app.mount so this is prettier.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -141,14 +141,18 @@ async def add_db_session(request: Request, call_next):
     return response
 
 async def get_context(request: Request):
-    
     dbsession = request.scope['dbsession']
     
     # allow for manual specification of user in request header by email
     if settings.debug and "X-User" in request.headers:
         user = User.get_by_email(session=dbsession, email=request.headers['X-User'])
     else:
-        user = await fastapi_users.current_user(active=True, optional=True)()
+        # TODO(jnu): fastapi_users.current_user is meant to be called with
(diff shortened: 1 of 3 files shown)


seerocode (approved):

api/app.py
```diff
@@ -141,14 +141,18 @@ async def add_db_session(request: Request, call_next):
...
     # allow for manual specification of user in request header by email
     if settings.debug and "X-User" in request.headers:
         user = User.get_by_email(session=dbsession, email=request.headers['X-User'])
     else:
-        user = await fastapi_users.current_user(active=True, optional=True)()
+        # TODO(jnu): fastapi_users.current_user is meant to be called with
+        # FastAPI's `Depends`. We have to hook into their "blood magic" here
+        # to call it outside of Depends.
+        user_db = await fastapi_users.current_user(active=True, optional=True)(cookie=request.cookies.get('rtauth'))
+        # TODO(jnu): The users.UserDB and database.User model should be unified.
```
nmanduj1, line 154: Might it make sense to take care of some of these todo's upfront? Mainly unifying the users.UserDB and database.User models?

jnu: my concern with that is that integrating with the sqlalchemy ORM was deliberately left out by the fastapi-users developers: https://frankie567.github.io/fastapi-users/configuration/databases/sqlalchemy.html#what-about-sqlalchemy-orm

i'm looking through their code now to see if i see a reasonable easy way to do it

PR #76, opened 2021-06-17
Add admin routes with placeholders and sidebar

jnu:
- Add an admin route list that is not available to normal users. Normal users will see a "not authorized" message when trying to view; non-logged-in users will be redirected to login screen.
 - Customize sidebar for admins to give access to the different sections, per mocks
 - Add placeholders for the three main management pages: teams, users, programs

 To do (later, not in this PR):
 - Main "view all datasets" view query needs to be updated for admins to see a full list of datasets. We could also just hide this for now.
 - Implement management page components to replace Todos
 - Add more routes like `/admin/users/<id>` to allow editing an individual entity.

image
using dark theme to differentiate from the normal users view, but i could go either way.


seerocode:
I like the idea of a different color for the sidebar to show what's an admin view vs a user view. Maybe we can explore a different color theme than dark mode but okay for now.

client/public/locales/en-gb/translation.json
```diff
@@ -11,5 +11,6 @@
   "addData": "Add Data",
   "editData": "Edit",
   "backToDashboard": "Back To Dashboard",
-  "noDataAvailable": "No Data Available"
+  "noDataAvailable": "No Data Available",
+  "adminSidebarManagePrograms": "Programmes"
```
seerocode, line 15: We can use reuse the `"program"` key instead where adminSidebarManagePrograms is required.

seerocode (approved):

PR #77, opened 2021-06-18
User management API routes

jnu:
Revises the fastapi-users sqlalchemy integration to be compatible with ORM.

still left to do:
 - [x] restrict permissions on register route
 - [x] tests

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -138,7 +138,7 @@ async def add_user(request: Request, call_next):
     if settings.debug and "X-User" in request.headers:
         user = User.get_by_email(session=dbsession, email=request.headers['X-User'])
     else:
-        # TODO(jnu): fastapi_users.current_user is meant to be called with
+        # NOTE(jnu): fastapi_users.current_user is meant to be called with
         # FastAPI's `Depends`. We have to hook into their "blood magic" here
         # to call it outside of Depends.
         user_db = await fastapi_users.current_user(active=True, optional=True)(cookie=request.cookies.get('rtauth'))
--- a/api/app.py
+++ b/api/app.py
@@ -3,9 +3,7 @@ import databases
 import sqlalchemy
 import datetime
 
-from fastapi import FastAPI, Request, Depends
-# from fastapi_sqlalchemy import DBSessionMiddleware  # middleware helper
-# from fastapi_sqlalchemy import db  # an object to provide global access to a database session
+from fastapi import FastAPI, Request, Depends, HTTPException
 from fastapi_users.authentication import CookieAuthentication
 from fastapi_users import FastAPIUsers
 from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base
@@ -42,24 +40,35 @@ cookie_authentication = CookieAuthentication(
 fastapi_users = FastAPIUsers(
     user.user_db,
     [cookie_authentication],
-    user.User,
-    user.UserCreate,
-    user.UserUpdate,
-    user.UserDB,
+    user.UserModel,
+    user.UserCreateModel,
+    user.UserUpdateModel,
+    user.UserDBModel,
 )
 
 
 # Add login functions 
-def on_after_register(user: user.UserDB, request: Request):
+def on_after_register(user: user.UserDBModel, request: Request):
     print(f"User {user.id} has registered.")
 
 
-def on_after_forgot_password(user: user.UserDB, token: str, request: Request):
+def on_after_forgot_password(user: user.UserDBModel, token: str, request: Request):
     print(f"User {user.id} has forgot their password. Reset token: {token}")
 
 
-def after_verification_request(user: user.UserDB, token: str, request: Request):
+def after_verification_request(user: user.UserDBModel, token: str, request: Request):
     print(f"Verification requested for user {user.id}. Verification token: {token}")
+
+
+def admin_user(request: Request):
+    """Dependency to verify that a user is an admin."""
+    user = request.scope.get('dbuser')
+    if not user:
+        raise HTTPException(status_code=401, detail="You are not authenticated")
+    if not any(r.name == 'admin' for r in user.roles):
+        raise HTTPException(status_code=403, detail="You do not have permission for this action")
+    return user
+    
 
 
 # Add restful routers for user management
@@ -67,7 +76,10 @@ app.include_router(
     fastapi_users.get_auth_router(cookie_authentication), prefix="/auth/cookie", tags=["auth"]
 )
 app.include_router(
-    fastapi_users.get_register_router(on_after_register), prefix="/auth", tags=["auth"]
+    fastapi_users.get_register_router(on_after_register),
+    prefix="/auth",
+    dependencies=[Depends(admin_user)],
+    tags=["auth"],
 )
 app.include_router(
     fastapi_users.get_reset_password_router(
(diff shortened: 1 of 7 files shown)


api/app.py
```diff
@@ -42,32 +40,46 @@
...
 
-def after_verification_request(user: user.UserDB, token: str, request: Request):
+def after_verification_request(user: user.UserDBModel, token: str, request: Request):
     print(f"Verification requested for user {user.id}. Verification token: {token}")
 
 
+def admin_user(request: Request):
+    """Dependency to verify that a user is an admin."""
+    user = request.scope.get('dbuser')
+    if not user:
```
seerocode, line 66: Which endpoint would be throwing these exceptions? Wondering so we could catch on the front end for these different codes.

api/app.py
```diff
@@ -42,32 +40,46 @@
...
+        raise HTTPException(status_code=401, detail="You are not authenticated")
+    if not any(r.name == 'admin' for r in user.roles):
```
seerocode, line 68: This is a loop because a user can have multiple roles?

api/app.py
```diff
@@ -129,6 +129,26 @@ def parse_datetime(value: str) -> datetime.datetime:
...
+
+@app.middleware("http")
+async def add_user(request: Request, call_next):
+    dbsession = request.scope["dbsession"]
+
+    # allow for manual specification of user in request header by email
+    if settings.debug and "X-User" in request.headers:
+        user = User.get_by_email(session=dbsession, email=request.headers['X-User'])
+    else:
+        # TODO(jnu): fastapi_users.current_user is meant to be called with
```
seerocode: What would be the next step for this TODO?

api/app_test.py
```diff
@@ -24,6 +27,16 @@
 from uuid import UUID
 
 
+def user_cookie(user_id):
+    data = {"aud": "fastapi-users:auth"}
+    if user_id:
```
seerocode, line 32: What happens if there isn't a user ID?

api/user.py
```diff
@@ -1,30 +1,99 @@
...
+    name: str
+    description: str
+
+    class Config:
+        orm_mode = True
+
+
+class UserModel(models.BaseUser):
     first_name: str
     last_name: str
```
seerocode, line 49: Oh, cool. So we can also add other attributes in the future if needed to the user model?

api/user.py
```diff
@@ -1,30 +1,99 @@
...
     last_name: Optional[str]
+    roles: Optional[List[str]]
 
 
-class UserDB(User, models.BaseUserDB):
-    # NOTE(jnu): the database.User model is much more useful than this one.
-    # We should figure out how to unify them.
+class UserDBModel(UserModel, models.BaseUserDB):
     first_name: str
     last_name: str
```
seerocode, line 72: Why does this model not include roles like the others?

api/user.py
```diff
@@ -1,30 +1,99 @@
...
+    Based on SQLAlchemyUserDatabase:
+    https://github.com/frankie567/fastapi-users/blob/c83bdeb0e0004692f398ad89a4fdeaaa125900d5/fastapi_users/db/sqlalchemy.py#L92
+    """
+
+    def __init__(self, session_factory):
+        super().__init__(UserDBModel)
+        self.session_factory = session_factory
+
+    async def get(self, id: UUID4) -> UserDBModel:
+        # TODO! Use an async session
```
seerocode, line 96: What's the level of effort to make this async?

seerocode: Also what happens if they're not async?

client/src/services/auth.ts
```diff
@@ -1,3 +1,11 @@
+/**
+ * Description of a role assigned to a user.
+ */
+export type UserRole = Readonly<{
```
seerocode, line 4: Awesome! I haven't made use of Readonly yet. Is this for safety reasons?

client/src/__tests__/auth.test.tsx
```diff
@@ -1,7 +1,9 @@
 import { mockUserLoggedIn, mockUserLogIn } from "../__mocks__/auth";
 
 it("marks user as admin correctly", async () => {
-  const { auth } = mockUserLoggedIn({ roles: ["admin"] });
+  const { auth } = mockUserLoggedIn({
```
seerocode, line 4: Oh cool. I missed the mockUserLoggedIn function before.

seerocode (approved):

api/app.py
jnu: this is the /auth/register route that will be called in the create user page. i have another branch that handles these errors. generally we can just handle anything that's not 200 or 201 should show an error message -- we could add special handling for 401 to redirect to the login page. (in practice 403 should never come up in ordinary usage.)

api/app.py
jnu: yep! we only have one defined now but we've discussed adding others.

api/app.py
jnu: hmm, i guess i really meant `NOTE` or something. i don't think there's a real TODO.

api/app_test.py
jnu: any protected routes will return "401 - not authorized"

api/user.py
jnu: yea, it's still a little funny because for the most part it's better to use GraphQL for everything. the amount of effort needed to bend the fastapi-users library to fit our use case is *almost* not worth it, but it does some nice stuff like password management and cookie validation that's good to keep.

api/user.py
jnu: it does! it inherits `UserModel` which has roles, so they don't need to be repeated down here. I've made a couple other slight modifications to this file in my edit users UI branch which i'm going to merge back in here now.

seerocode: Oh got it! I missed that.

api/user.py
jnu: great questions!

1) we need to upgrade to sqlalchemy 1.4 and rewrite our database queries using the new syntax in order to make async queries. estimating about an afternoon's worth of work.

2) there's a significant performance hit. in practice web servers like ours spend most of their time waiting around for network requests like database queries to return and relatively little time actually executing their own code. this is really inefficient to do in real OS-level threads, because OS threads are expensive to run -- lots of memory and in python it's slow to switch between them. async web servers are much more efficient because they let you handle a huge amount of web requests in one single thread -- instead of just waiting around for a network request like a database query to return (like an old threaded web server would do), they use that idle time to process other requests that are queued up, so they are able to juggle a bunch of requests at once -- these are called "green threads". modern async webservers like fastapi use both a small threadpool (usually matching the number of CPUs your computer has -- so small, like 4 or 8), and run these async green threads in each thread.

all that said, by **not** using an async database request here, we are blocking any other green thread from running in that OS thread.

using async, each OS thread is typically able to handle thousands of requests per second. if you use blocking IO like this, it can't come anywhere close to that number. if a db request takes 50ms to return, a thread that might've been able to process 5000 requests per second can now at best only do 20.

in practice for us, users likely will notice sluggish performance when we roll out to the entire org (~1000 people) if we keep the synchronous calls.

client/src/services/auth.ts
jnu: yea, generally i feel like you might as well make everything that comes from the API Readonly just for sanity, since any mutations made on the frontend won't be persisted in the backend.

PR #78, opened 2021-06-21
Ch274/all crud

nmanduj1:


api/database.py
```diff
@@ -455,6 +457,7 @@ def hash_test_password(pw: str) -> str:
            entries=[entry1, entry2, entry3, entry4, entry5, entry6, entry7, entry8]) 
     
     session.add(category_value_white)
+    session.add(testing_team)
```
seerocode: What's the different between team on line 413 and testing_team on line 416?

api/mutations.py
```diff
@@ -234,4 +234,62 @@ def resolve_delete_category_value(obj, info, id):
...
+        :param input: params for new Team
+        :returns: Team dictionary
+    '''
+
+    session = info.context['dbsession']
+    users = input.pop('user_ids')
+    programs = input.pop('program_ids')
+    
+    team = Team(**input)
+    team.programs += [session.merge(Program(id=program_id)) for program_id in programs]
```
seerocode, line 252: Probably not for now but I wonder if we should be adding inputter id for all crud so we have a log of who made a change where at any point in time?

api/queries.py
```diff
@@ -81,4 +81,15 @@ def resolve_category_value(obj, info, id):
...
+    return category_value
+
+@query.field("team")
+def resolve_team(obj, info, id):
+    '''GraphQL query to find a Team based on Team ID.
+        :param id: Id for the Team to be fetched 
+        :returns: Team dictionary 
+    '''
+    session = info.context['dbsession']
+    team = session.query(Team).get(id)
```
seerocode, line 133: Does this need to check if it's been deleted?

api/schema.graphql
```diff
@@ -46,7 +46,7 @@ type Team @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   name: String!
   users: [User!]!
   programs: [Program!]!
-  organizations: [Organization!]!
+  organization: Organization!
```
seerocode, line 49: Oh great catch!

api/schema.graphql
```diff
@@ -198,6 +198,21 @@ input UpdateCategoryValueInput {
   categoryId: ID
 }
 
+input CreateTeamInput {
+  name: String!
+  userIds: [ID!]!
```
seerocode, line 210: Does this mean users need to be added to a team whenever one is created?

api/schema.graphql
```diff
@@ -198,6 +198,21 @@ input UpdateCategoryValueInput {
...
+  programIds: [ID!]!
```
seerocode, line 211: Similarly here, can empty teams exist until there are programs and users assigned to them?

api/database.py
nmanduj1: The db does not allow a Team with associations to be deleted (in order to maintain data integrity). `testing_team` has no associations and is used for testing the `deleteTeam` mutation (which is a hard delete) though now that you point this out, i am missing a cascade to hard delete the associations in the `user_teams` table- please see comment below.

api/mutations.py
nmanduj1: We talked about adding an Audit table- that would be a better way to handle logging. Hopefully we can get to that soon!

api/queries.py
nmanduj1: Not currently because we are not soft deleting Teams so anything that exists in the Teams table with that team_id should be retrieved. Although now that you point this out, we might want Teams to also be soft-deleted (?) maybe? Idk that soft-deleting Teams would be useful but I can update this if you see value in that functionality. I will bring this up during eng-sync so we can decide how the Team deletes should be handled.

api/schema.graphql
nmanduj1: thank u :)

api/schema.graphql
nmanduj1: I would expect Users to exist and be added when a Team is created, but a Team can be created with no Users here by passing in an empty `userIds` array

api/schema.graphql
nmanduj1: Yes, one can pass in an empty `userIds` and `programIds` array.

api/schema.graphql
seerocode: Oh okay I read this as a non-empty array of non-null IDs is required.

api/schema.graphql
seerocode: Would it make sense to change it to `[ID!]` instead so an empty array isn't required?

nmanduj1: I think it makes sense to require an empty array when a Team is created to ensure it was intentionally left without being associated to programs / users. The arrays will not be required when the Team is being updated.

seerocode: Sounds good. That's easy enough to manage as an intention on the client side.

api/mutations.py
seerocode: I'll add the audit table as a story to groom.

seerocode (approved):
🚢

nmanduj1:
Okedokes, as we discussed during eng sync, I left the hard deletes for Team. 

Just a quick recap on functionality here: when a `Team` is hard deleted, the association in the `user_team` table is also hard deleted and the `team_id` column in the `Program` is set to `null` (Programs can then be associated with a different Team). 

Because of these changes, I was able to clean up the seed data a bit and I added additional test coverage to make sure the functionality trickles down to `user_team` and `Program` as we expect. Tests 💯 !

PR #79, opened 2021-06-23
[ch159] Search bar should filter table

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/159

### Changes:
- added search handler to search bar on home page to filter datasets table by search term
    - when no results are found, table will show all available data 
- added margin to breadcrumbs
 
### Screenshots:
filter


client/src/components/Home/Home.tsx
```diff
@@ -11,24 +11,37 @@ import localizedFormat from "dayjs/plugin/localizedFormat";
...
+  tags: Array<string>;
+}
+
+const columns: ColumnsType<TableData> = [
   {
     title: "Team",
     dataIndex: "team",
     key: "team",
+    sortDirections: ["ascend", "descend"],
+    sorter: (a, b) => a.team.localeCompare(b.team),
```
nmanduj1, line 33: Neat! Does this also sort numbers? I imagine this is based on ascii values?

client/src/components/Home/Home.tsx
```diff
@@ -91,8 +104,8 @@ const getTableData = (
           lastUpdated: dataset.lastUpdated
             ? dayjs(dataset.lastUpdated).format("ll")
             : t("noDataAvailable"),
-          tags: dataset.tags.map((t) => {
-            return t.name;
+          tags: dataset.tags.map((tag) => {
```
nmanduj1, line 107: I really appreciate this update, it makes this easier to follow- thank you!

client/src/components/Home/Home.tsx
```diff
@@ -113,30 +126,53 @@ const Home = (): JSX.Element => {
...
         <div>
-          <HomeSearchAutoComplete
-            dataSource={rowData.map(
-              (i: { team: string; dataset: string }) =>
-                `${i.team} - ${i.dataset}`
-            )}
-          />
+          <div
+            id="home_table-search"
+            style={{
```
nmanduj1, line 155: I am wondering if this Home component has a stylesheet? / Does it make sense to have its own stylesheet? Also, I saw the render bugs you identified yesterday. They seem to happen randomly for some reason. I am thinking we should probably invest some time into making sure the app is responsive.

client/src/components/Home/Home.tsx
```diff
@@ -113,30 +126,53 @@ const Home = (): JSX.Element => {
...
+            }}
+          >
+            <HomeSearchAutoComplete onSearch={handleTableSearchFilter} />
+          </div>
           <Table
-            dataSource={rowData}
+            dataSource={filteredData.length > 0 ? filteredData : rowData}
             columns={columns}
             rowKey={(dataset) => dataset.id}
+            footer={() =>
```
nmanduj1, line 167: Will this table footer always render, whether or not the user is using the search bar?

client/src/components/Home/HomeSearchAutoComplete.tsx
```diff
@@ -3,28 +3,26 @@ import { AutoComplete, Input } from "antd";
...
+interface AutoCompleteProps {
+  onSearch: (text: string) => void;
 }
 
-const HomeSearchAutoComplete = ({ dataSource }: Props): JSX.Element => {
-  const { t, i18n } = useTranslation();
+const HomeSearchAutoComplete = ({
+  onSearch,
+}: AutoCompleteProps): JSX.Element => {
+  const { t } = useTranslation();
```
nmanduj1, line 15: t denotes a translation here?

client/src/components/Home/Home.tsx
seerocode: Yeah and this can be moved to a sorter utility to sort by an enum like NUMBER or DEFAULT with default being anything other than numbers but I felt for now this was fine.

It's the same as `stringA > stringB ? 1 : -1` but this sorts by the user's OS locale so actually maybe we don't want to do that?

client/src/components/Home/Home.tsx
seerocode: Thanks! It was confusing to have t for the translation hook and t for tag representation. 🤦🏽

client/src/components/Home/Home.tsx
seerocode: I added a stylesheet for each component since Home itself is just a page for holding components. That's a structure preference though and we can change that. I think the structure of the frontend needs to change again to define "pages" and "components" separately now that we have re-usable components. 

This PR also fixes the table rendering issue where the background gets cut off!

client/src/components/Home/Home.tsx
seerocode: Yep it will so it will show n of n results to provide the number of datasets in the table view.

client/src/components/Home/HomeSearchAutoComplete.tsx
seerocode: yeah, it's a property within the useTranslation hook we are pulling out.

client/src/components/Home/Home.tsx
nmanduj1: I agree leaving it here is fine for now- we can adjust as necessary in the future!

nmanduj1 (approved):
🚢

PR #80, opened 2021-06-23
Nginx fix

nmanduj1:


.env
```diff
@@ -6,7 +6,7 @@
 # Load dummy data for demos / development when initializing the DB.
 # Set this to "1" to enabled.
 # TODO: change this to run arbitrary init scripts
-RT_DUMMY_DATA=0
```
seerocode, line 9: I think this needs to go back to zero.

api/app.py
```diff
@@ -35,7 +35,7 @@
 cookie_authentication = CookieAuthentication(
         secret=settings.secret,
         lifetime_seconds=3600,
-        cookie_secure=not settings.debug,
```
seerocode, line 38: What does setting to false do?

nginx.conf
```diff
@@ -50,5 +50,19 @@ http {
             proxy_set_header Host $http_host;
             proxy_pass http://api$uri;
         }
+
+        location /users/me {
```
seerocode, line 54: 🥇 for the nginx fix!!!

api/app.py
nmanduj1: We do not want this to be false in production- this is just a temp fix to get things running for the BBC Demo tomorrow. As this was set, the cookie was secure and the browser was rejecting it because it was not being transmitted over https- essentially our certificates were not inline.

.env
nmanduj1: hm, definitely works set to 1, not sure if it will work being set to 0- theoretically? but theoretically led us into today- i can look into it tomorrow. Or you can check in if you'd like to play with it today

seerocode: I think zero was the default so BBC and other partner sides can have their own data. Not sure how they've got it working without the flag on their end though!

seerocode (approved):

api/app.py
seerocode: Let's throw a ticket in clubhouse for this then!

nginx.conf
jnu: woo!! awesome work debugging this issue and putting up a fix!!! 🤩

PR #81, opened 2021-06-29
[ch226] Publication date does not clear after user selects "save add new record" on the Add Entry page

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/226

Changes:
- added test for this issue
- fixed issue with date field not clearing after clicking the second save button by using refs
fix
- added generic wrappers to Apollo mutation hooks to make reuse easier


client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -92,23 +90,26 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
...
-  ] = useMutation(CREATE_RECORD, {
-    awaitRefetchQueries: true,
-    refetchQueries: [
-      {
-        query: GET_DATASET,
-        variables: { id: props.datasetId },
-      },
-    ],
-  });
+  const isSaveAndAddAnotherRecordClicked = useRef(false);
```
seerocode: Since state is asynchronous, using a state hook was causing a race condition and making the test for this fix fail because there was no guarantee we could immediately get the state result back once it was changed.

Refs are synchronous so the next two lines are sort of functioning similar to a `useState` hook in its implementation of a value retrieval  (`isSaveAndAddAnotherRecordClicked`) and a setter (`setIsSaveAndAddAnotherRecord`). 

Not sure if there's a more appropriate way to set this up so I'm interested in any other ideas!

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -278,9 +269,12 @@ const DataEntryAggregateDataEntryForm = (props: FormProps): JSX.Element => {
             {isEditMode ? (
               <Button
                 type="primary"
+                name="save"
+                value="update record"
                 htmlType="submit"
                 icon={<SaveOutlined />}
                 style={{ whiteSpace: "normal", height: "auto" }}
+                onClick={() => setIsSaveAndAddAnotherRecord(false)}
```
seerocode: Kind of bugs me that I need to set this in every submit button. Thoughts?

client/src/components/DataEntry/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+import { GetRecord } from "../../__generated__/GetRecord";
+import { UpdateRecord } from "../../__generated__/UpdateRecord";
+
+type CustomMutationHook<T, R extends MutationResult> = (input: T) => R;
+
+interface CreateRecordMutationProps {
+  datasetId: string;
+}
+
+interface UseCreateRecordResult extends MutationResult<GetRecord> {
```
seerocode: Is there a way I can make the property of this interface generic so I don't have to write this for each wrapper?

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
jnu: huh. interesting idea! definitely a bit of a hack, since refs are designed to hold references to DOM elements. the "expected" way of using a ref here would be something like: create two refs to hold references to the DOM elements of each submit button. then in the onSubmit handler, check the `submitter` on the submit event against each refs `current` DOM element to figure out which one was clicked.

jnu: i suppose there's a way to do this with useState and useEffect as well:

```ts
const [submit, setSubmit] = useState(false);
const [whichButton, setWhichButton] = useState<'' | 'normal' | 'other'>('');

useEffect(() => {
  if (!submit || !whichButton) {
    // resolve the race condition by waiting for both values to be set
    return;
  }

  if (whichButton === 'normal') {
    // do normal submit stuff
  } else if (whichButton === 'other') {
    // do other submit stuff
  }

  // reset state
  setSubmit(false);
  setWhichButton('');
// Condition on both state values to run this effect:
}, [submit, whichButton]);


// then you can use both onSubmit and onClick:
<form onSubmit={() => setSubmit(true)}>
  <button type="submit" onClick={() => setWhichButton("normal") value="Normal submit" />
  <button type="submit" onClick={() => setWhichButton("other") value="Other submit" />
</form>
```

i haven't tested that, but i think it should work!

client/src/components/DataEntry/hooks.ts
jnu: which property? `createRecord`? looks like it's only used in one place, unless i'm missing something?

client/src/components/Error/ErrorFallback.tsx
```diff
@@ -12,10 +12,11 @@ export const ErrorFallback = ({ error }: ErrorProps): JSX.Element => {
     <Alert
       className="error-alert"
       role="alert"
-      message="Something went wrong"
+      message="Oh, no! Something went wrong"
```
jnu, line 15: no emoji? 😛

jnu (approved):
looks fantastic! i made a comment about the refs. i guess if it works it works, i'm just nervous that "refs" weren't really built for this. do you know if the synchronousness is a "guaranteed" feature of the refs API? my gut is that it's better not to rely explicitly on the timing of anything in the render cycle, since i can imagine it all being changed to optimize for render performance.

if you do keep it as is, could u please add a comment like the one you wrote on GH to the code to explain whats going on and why?

in any case, great job with finding a creative solution!!! this is a really tricky problem and i'm also not sure the "right" way to solve it!

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
seerocode: I'm going to try this out because I don't like the way refs are set up here either. Feels too close to a state hook which like you mentioned, isn't what they're for so I would prefer not to hack it like that. Although I was using one ref for both buttons so that could have been the first issue! Thanks for driving me in a better direction.

client/src/components/DataEntry/hooks.ts
seerocode: Yep, `createRecord`. The interface for this hook is almost exactly the same as the second one mutation hook in this file. I thought maybe there was a way to reuse an interface an instead?

client/src/components/Error/ErrorFallback.tsx
seerocode: 😆 We can if we want to! It has a plain old X "error" icon by default.

client/src/components/DataEntry/DataEntryAggregateDataEntryForm.tsx
seerocode: So Safari doesn't support the "submitter" field for a synthetic event so we can't use that with refs to check what button was clicked. Using an effect also doesn't seem to do the same as "componentDidUpdate" as I had hoped to make the test not cry.

PR #84, opened 2021-07-02
Admin UI: user management

jnu:
Admin UI for creating, listing, editing, and deleting users.

in this pr
 - [x] create new API
 - [x] edit user UI
 - [x] list all users UI
 - [x] merge api changes into #77 
 - [x] forgot password UI flow
 - [x] verification email UI handler
 - [x] email send handler for user registration / forgot password / verification
 - [x] tests for fastapi-users routes
 - [x] tests for graphql schema changes
 - [x] Accessibility questions about Forms
 - [x] UI polish
 - [x] UI tests


jnu:
screens to get a gist of the UI changes:

==admin screens==

User list:
image

Create user [clicking save redirects to the Edit user page below. I don't love the UI, but it fits most easily with how the fastapi-users API was designed (e.g., you can't assign admin privileges when you create a user, only when editing)]:
image

Edit a user:
image

Assign to a team:
image

Deactivation modal:
image

Editing inactive user / reactivating:
image

==general user screens==

Forgot password on login:
image

Email confirmation:
image

Reset password screen [the token in the URL authorizes this request, then redirects to login on success]:
image

Error if the token is invalid:
image

api/app.py
```diff
@@ -88,23 +98,43 @@ def admin_user(request: Request):
...
+# I submitted a pull request to fix this in fastapi-users here:
+# https://github.com/frankie567/fastapi-users/pull/650
+# 
+# More info here:
+# https://github.com/tiangolo/fastapi/issues/717
+#
+# When the PR is merged, bump the fastapi-users version and remove the hack.
+# For now, reach into the router's delete_user route and set the response class
+# explicitly to the bare Response class to avoid issues.
+users_router = fastapi_users.get_users_router()
```
seerocode, line 127: PR has been merged so this can be changed now: https://github.com/frankie567/fastapi-users/pull/650#issuecomment-873448596

api/mailer.py
```diff
@@ -0,0 +1,143 @@
...
+    :param body: Email body as HTML
+    :returns: New email object
+    """
+    email = EmailMessage()
+    email["From"] = Address(
+            display_name=settings.email.send_from_display_name,
+            addr_spec=settings.email.send_from_address)
+    email["To"] = to
+    email["Subject"] = subject
+    email.set_content(render_html_as_text.handle(body))
```
seerocode, line 64: Do you have a screenshot of what this looks like when it's converted from HTML to ascii text?

api/queries.py
```diff
@@ -22,10 +23,24 @@ def resolve_user(obj, info, id):
...
     session = info.context['dbsession']
-    user = session.query(User).filter(User.id == id, User.deleted == None).first()
+    user = session.query(User).filter(User.id == id).first()
 
     return user
 
+@user.field("active")
+def resolve_user_active(user, info):
+    '''Field indicating whether user is currently active.'''
+    return user.deleted is None
```
seerocode, line 33: The users query will now return all users whether they were deleted or not and an "active" field to check their status in the User object?

api/queries.py
```diff
@@ -127,9 +142,27 @@ def resolve_category_value(obj, info, id):
...
     return team
+
+
+@query.field("teams")
+def resolve_teams(obj, info):
+    '''GraphQL query to fetch full list of teams.
+    :returns: List of team objects
+    '''
+    session = info.context['dbsession']
+    return session.query(Team).filter(Team.deleted == None).order_by(Team.name.asc()).all()
```
seerocode, line 159: We could probably add this delete check to the database class for consistency.

api/settings.py
```diff
@@ -1,4 +1,84 @@
...
+
+
+class SmtpSettings(BaseModel):
+    # Turn on the mock to skip sending real emails. Handy for development.
+    # Turn this off in production!
+    mock: bool = True
+    # SMTP host
+    host: str = "localhost"
+    # SMTP port on the remote server
+    port: int = 25
```
seerocode, line 11: Do we need to handle this port at all in the Docker container?

client/src/components/Admin/CreateUser.tsx
```diff
@@ -0,0 +1,97 @@
...
+        message.warn(e.message);
+      }
+      history.push(`/admin/users/${id}`);
+    } catch (e) {
+      message.error(e.message);
+    }
+  };
+
+  return (
+    <Form
```
seerocode: Can we check the accessibility on these forms? I had used a native HTML form for the data input page because there were accessibility issues raised in devtools when I tried to use a AntD form. Hopefully it's been fixed.

client/src/components/Admin/UserList.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { data, loading, error } = useQuery<GetUserList>(GET_USER_LIST, {
+    fetchPolicy: "network-only",
+  });
+
+  // TODO: update error and loading components
+  if (loading) {
+    return <Loading />;
+  }
+
+  if (error || !data || !data.users) {
```
seerocode: We can use the `ErrorFallback` component here. It's the AntD Error component.

client/src/components/Admin/UserList.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return <div>An error occurred: {error}</div>;
+  }
+
+  const columns: ColumnsType<GetUserList_users> = (
+    ["email", "firstName", "lastName"] as Array<keyof GetUserList_users>
```
seerocode: This is saying these array items are keys of `GetUserList_users`?

client/src/components/Admin/UserList.tsx
```diff
@@ -0,0 +1,111 @@
...
+  ).map((c) => ({
```
seerocode: Can `c` be renamed to something more descriptive? I think I did this too in some places and ended up renaming in subsequent PRs.

client/src/components/VerifyAccount.tsx
```diff
@@ -0,0 +1,105 @@
...
+        setVerifyError(e);
+      })
+      .then(() => {
+        setVerifying(false);
+      });
+  }, []);
+
+  // Verification should be very quick, but show a spinner just in case.
+  if (verifying) {
+    return <Loading />;
```
seerocode: This is a full-page spinner? If so, maybe we do the skeleton loading instead of a spinner? Fine for now.

client/src/router/routes.tsx
```diff
@@ -80,9 +82,16 @@ export const adminRoutes: IRoute[] = [
...
-        component: Todo("manage users"),
+        component: UserList,
         breadcrumb: "Manage users",
       },
+      {
+        path: "/admin/users/:userId",
+        key: "user",
+        exact: true,
+        component: EditUser,
+        breadcrumb: "Edit User",
```
seerocode, line 93: Small nit! The casing is different here than the rest of the breadcrumbs.

client/src/services/account.ts
```diff
@@ -0,0 +1,248 @@
...
+}>;
+
+/**
+ * Data that's generated by the UI form to edit the user.
+ */
+export type EditUserFormData = Readonly<{
+  first_name: string;
+  last_name: string;
+  email: string;
+  roles: string[];
```
seerocode, line 17: Should we have an interface for roles?

client/src/services/account.ts
```diff
@@ -0,0 +1,248 @@
...
+  const r = await fetch("/auth/verify", {
+    method: "POST",
+    credentials: "same-origin",
+    body: JSON.stringify({ token }),
+    headers: [["Content-Type", "application/json"]],
+  });
+
+  let json;
+
+  switch (r.status) {
```
seerocode: Might be good to move this to a global error handler in the future.

api/settings.py
jnu: good question! this is an outbound connection (our web app sends data to the specified SMTP server, which mails the message), which should be permitted by default.

i added a chore to test verify that this all works with a real smtp server, in part because the rendered HTML also sometimes looks funky:
https://app.clubhouse.io/stanford-computational-policy-lab/story/307/test-sending-real-emails-via-smtp

client/src/components/VerifyAccount.tsx
jnu: what do you mean by skeleton loading?

client/src/services/account.ts
jnu: what do you mean by interface? the type at this point is a string (just the role ID).

client/src/services/account.ts
jnu: the error responses are somewhat different based on the endpoint, but there may be some common stuff we could factor out

api/mailer.py
jnu: it basically just adds paragraph breaks:

image

for context, email is an old protocol and by default emails are plain text, without formatting. emails these days are generally sent as "multipart" messages that specify both a text/plain version for backwards-compatibility and a text/html version for nice formatting. nobody ever sees the text/plain version of an email unless they explicitly turn off html formatting in their email client.

api/app.py
jnu: unfortunately we should wait for a new tagged release version before removing. you _can_ specify a git sha in requirements.txt, but you run the risk of pulling intermediate / broken changes, so it's better to wait for the maintainers to ship a release. since the workaround is small i don't mind waiting.

api/queries.py
jnu: that's correct. the `users` query is new so worth noting i didn't change any behavior there. the `user` (singular) query I modified to return inactive users to support the "reactivate" button.

api/queries.py
jnu: the non-classmethods actually look like the norm in this file, so i'd say it's consistent! :p 

i agree we should pick a strategy and stick with it. worth nothing we have to rewrite all these queries to support https://app.clubhouse.io/stanford-computational-policy-lab/story/302/use-async-db-calls so i think it's ok to punt for now.

client/src/components/Admin/CreateUser.tsx
jnu: interesting, i will check accessibility. i realized as i was writing this that the issues you were having with the submit behavior were likely due to mixing the native html form with the Input elements from antd. antd also has a well-developed system of form validation (and UI layout, including labels and error messages) that we'd miss out on if we dropped their Form component.

i'll investigate the accessibility issues -- i'd much rather find a workaround and/or fix the issues upstream (as i did for the fastapi-users) than not be able to use their components.

client/src/components/Admin/UserList.tsx
jnu: yep! i specified this so the `sorter` function could be typechecked properly. typescript will also complain if you put items in that list that _aren't_ keys of that users object.

api/app.py
seerocode: Probably better to include as a story then since it may be easy to forget.

client/src/components/Admin/UserList.tsx
jnu: ah good to know, thanks! fwiw i saw some TODOs about loading/errors in other files that i essentially copied here. will update.

seerocode: Yeah :/ I've been cleaning them up as I go. It would be better to have like an error wrapper but there are issues with async stuff there.

client/src/components/VerifyAccount.tsx
seerocode: https://ant.design/components/skeleton/

client/src/services/account.ts
seerocode: There was another place where roles was defined as either admin or something else. I can't find where it was now.

jnu: ah, you mean a string enum? like `type Role = 'admin';`?

for this part of the UI i wanted to make it fairly general -- we just request a list of all defined roles from the server (in case we add more later) and the UI uses their IDs (not their names) to toggle them on and off. so this "edit user" stuff doesn't know anything in particular about what any of the roles are. the IDs are just UUIDs, which are arbitrary.

in contrast, for the `auth` part of the UI it makes sense to check the `admin` role name explicitly to check if the user has that particular role.

client/src/components/VerifyAccount.tsx
jnu: ah i see, maybe i'm not sure what the skeleton would indicate though? there's not actually any UI on this page that's getting loaded -- in the normal case all this page does is redirect the user somewhere else. The flow is:

1) User gets an email requesting they verify their email by clicking the given link
2) User navigates to that link
3) That link sends the token to the backend for verification
4) Assuming all goes well they get redirected pretty much immediately

seerocode: Got it! The current spinner makes sense here then.

api/app.py
jnu: great idea, tracked here:
https://app.clubhouse.io/stanford-computational-policy-lab/story/308/remove-hack-for-fastapi-users-delete-handler

seerocode (approved):

client/src/components/Admin/UserList.tsx
jnu: sure, i'm always super lazy about iteration variable names :(

PR #85, opened 2021-07-07
[ch282] Add person type

seerocode:
Story details: 
- https://app.clubhouse.io/stanford-computational-policy-lab/story/282
- https://app.clubhouse.io/stanford-computational-policy-lab/story/283/update-ui-to-support-basic-entry-by-person-type

### Changes:
- Added person type to category sections in entry form if person type exists
image
- Added `personType` to `Entry` and `Dataset` as array (no new table for now so it is not in normal form)

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -153,6 +153,20 @@ program_tags = Table('program_tag', Base.metadata,
                          'tag.id'), index=True),
                      )
 
+dataset_person_types = Table('dataset_person_type', Base.metadata,
+                            Column('dataset_id', GUID, ForeignKey(
+                                'dataset.id'), index=True),
+                            Column('person_type_id', GUID, ForeignKey(
+                            'person_type.id'), index=True),
+                        )
+
+entry_person_types = Table('entry_person_type', Base.metadata,
+                            Column('entry_id', GUID, ForeignKey(
+                                'entry.id'), index=True),
+                            Column('person_type_id', GUID, ForeignKey(
+                            'person_type.id'), index=True),
+                        )
+
 
 class PermissionsMixin:
     """Base class defining some common permissions checks."""
@@ -515,13 +529,14 @@ class Dataset(Base, PermissionsMixin):
     __tablename__ = 'dataset'
 
     id = Column(GUID, primary_key=True, index=True, default=uuid.uuid4)
-    person_types = Column(ARRAY(String(255)))
     name = Column(String(255), nullable=False)
     description = Column(String(255), nullable=False)
     program = relationship('Program', back_populates='datasets')
     program_id = Column(GUID, ForeignKey('program.id'), index=True)
     records = relationship('Record')
     tags = relationship('Tag', secondary=dataset_tags,
+                        back_populates='datasets')
+    person_types = relationship('PersonType', secondary=dataset_person_types,
                         back_populates='datasets')
 
     # Datasets must be unique within a program to avoid ambiguity.
@@ -590,7 +605,8 @@ class Entry(Base, PermissionsMixin):
     record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
     inputter = relationship('User')
     inputter_id = Column(GUID, ForeignKey('user.id'), index=True)
-    person_type = Column(String(255), nullable=True)
+    person_types = relationship('PersonType', back_populates='entries')
+    person_type_id = Column(GUID, ForeignKey('person_type.id'), index=True)
 
     created = Column(TIMESTAMP,
                      server_default=func.now(), nullable=False)
@@ -608,6 +624,23 @@ class Entry(Base, PermissionsMixin):
     def soft_delete(self, session):
         self.deleted= func.now()
         session.add(self)
+
+class PersonType(Base, PermissionsMixin):
+    __tablename__ = 'person_type'
+
+    id = Column(GUID, primary_key=True, default=uuid.uuid4)
+    person_type_name = Column(String(255), nullable=False)
+    description = Column(String(255), nullable=True)
+    datasets = relationship('Dataset', secondary=dataset_person_types,
+                            back_populates='person_types')
+    entries = relationship('Entry', secondary=entry_person_types,
+                        back_populates='person_types')
+
+    created = Column(TIMESTAMP,
+                     server_default=func.now(), nullable=False)
+    updated = Column(TIMESTAMP,
+                     server_default=func.now(), onupdate=func.now())
+    deleted = Column(TIMESTAMP)
 
 
 def create_tables(session):
@@ -697,9 +730,9 @@ def create_dummy_data(session):
     session.add(admin_user)
     
     ds1 = Dataset(id='b3e7d42d-2bb7-4e25-a4e1-b8d30f3f6e89', name='Breakfast Hour',
-            description='breakfast hour programming', person_types=['BBC Contributor', 'Non-BBC Contributor'])
+            description='breakfast hour programming')
     ds2 = Dataset(id='96336531-9245-405f-bd28-5b4b12ea3798', name='12PM - 4PM',
-            description='afternoon programming', person_types=[])
+            description='afternoon programming')
     
     program = Program(id="1e73e788-0808-4ee8-9b25-682b6fa3868b", name='BBC News',
             description='All BBC news programming', datasets=[ds1, ds2])
@@ -724,22 +757,22 @@ def create_dummy_data(session):
     target_disabed = Target(id='b5be10ce-103f-41f2-b4c4-603228724993', program=program, target_date=datetime.strptime('2022-12-31 00:00:00', '%Y-%m-%d %H:%M:%S'), target=float(.50))
     target_non_disabled = Target(id='6e6edce5-3d24-4296-b929-5eec26d52afc', program=program, target_date=datetime.strptime('2022-12-31 00:00:00', '%Y-%m-%d %H:%M:%S'), target=float(.50))
 
-    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', count=1, inputter=user, person_type='BBC Contributor') 
-    entry2 = Entry(id='a37a5fe2-1493-4cb9-bcd0-a87688ffa409', count=1, inputter=user, person_type='BBC Contributor') 
-    entry3 = Entry(id='423dc42f-4628-40e4-b9cd-4e6e9e384d61', count=1, inputter=user, person_type='BBC Contributor') 
-    entry4 = Entry(id='407f24d0-c5eb-4297-9495-90e325a00a1d', count=1, inputter=user, person_type='BBC Contributor') 
-    entry5 = Entry(id='4adcb9f9-c1eb-41ba-b9aa-ed0947311a24', count=1, inputter=user, person_type='BBC Contributor') 
-    entry6 = Entry(id='1c49c64f-51e6-48fe-af10-69aaeeddc55f', count=1, inputter=user, person_type='BBC Contributor') 
-    entry7 = Entry(id='335b3680-13a1-4d8f-a917-01e1e7e1311a', count=1, inputter=user, person_type='BBC Contributor')
-    entry8 = Entry(id='fa5f1f0e-d5ba-4f2d-bdbf-819470a6fa4a', count=1, inputter=user, person_type='BBC Contributor')
-    entry1_b = Entry(id='9a323df8-15cf-45c2-a8c6-755b7d98332b', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry2_b = Entry(id='4c672b25-02da-4d48-bece-26b45fff9a03', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry3_b = Entry(id='21d12404-a1b7-40c0-9598-5dc853adbb9b', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry4_b = Entry(id='98aecacd-9eea-4d08-9b0d-a80c57db3b74', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry5_b = Entry(id='0ef0e83a-81b2-4c05-82d5-2fc5d8714368', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry6_b = Entry(id='4f07ef7f-70ec-4c04-9f57-fd692ab430d2', count=1, inputter=user, person_type='Non-BBC Contributor') 
-    entry7_b = Entry(id='201d163a-873c-4196-9056-18e66eab37c7', count=1, inputter=user, person_type='Non-BBC Contributor')
-    entry8_b = Entry(id='d7f57989-cf6e-4384-93d4-773d71137e0d', count=1, inputter=user, person_type='Non-BBC Contributor')
+    entry1 = Entry(id='64677dc1-a1cd-4cd3-965d-6565832d307a', count=1, inputter=user) 
+    entry2 = Entry(id='a37a5fe2-1493-4cb9-bcd0-a87688ffa409', count=1, inputter=user) 
+    entry3 = Entry(id='423dc42f-4628-40e4-b9cd-4e6e9e384d61', count=1, inputter=user) 
+    entry4 = Entry(id='407f24d0-c5eb-4297-9495-90e325a00a1d', count=1, inputter=user) 
+    entry5 = Entry(id='4adcb9f9-c1eb-41ba-b9aa-ed0947311a24', count=1, inputter=user) 
+    entry6 = Entry(id='1c49c64f-51e6-48fe-af10-69aaeeddc55f', count=1, inputter=user) 
+    entry7 = Entry(id='335b3680-13a1-4d8f-a917-01e1e7e1311a', count=1, inputter=user)
+    entry8 = Entry(id='fa5f1f0e-d5ba-4f2d-bdbf-819470a6fa4a', count=1, inputter=user)
+    entry1_b = Entry(id='9a323df8-15cf-45c2-a8c6-755b7d98332b', count=1, inputter=user) 
+    entry2_b = Entry(id='4c672b25-02da-4d48-bece-26b45fff9a03', count=1, inputter=user) 
+    entry3_b = Entry(id='21d12404-a1b7-40c0-9598-5dc853adbb9b', count=1, inputter=user) 
+    entry4_b = Entry(id='98aecacd-9eea-4d08-9b0d-a80c57db3b74', count=1, inputter=user) 
+    entry5_b = Entry(id='0ef0e83a-81b2-4c05-82d5-2fc5d8714368', count=1, inputter=user) 
+    entry6_b = Entry(id='4f07ef7f-70ec-4c04-9f57-fd692ab430d2', count=1, inputter=user) 
+    entry7_b = Entry(id='201d163a-873c-4196-9056-18e66eab37c7', count=1, inputter=user)
+    entry8_b = Entry(id='d7f57989-cf6e-4384-93d4-773d71137e0d', count=1, inputter=user)
     
     CategoryValue(id='6cae6d26-97e1-4e9c-b1ad-954b4110e83b', name='non-binary', category=category_gender, targets=[target_non_binary], entries=[entry1, entry1_b])
     CategoryValue(id='742b5971-eeb6-4f7a-8275-6111f2342bb4', name='cisgender women', category=category_gender, targets=[target_cis_women], entries=[entry2, entry2_b])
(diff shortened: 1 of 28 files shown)


seerocode:
The styling for the titles makes me 😞 but I wanted to get this out there for now and we can hash out what the styles should be!

api/database.py
```diff
@@ -384,6 +384,7 @@ class Dataset(Base, PermissionsMixin):
     __tablename__ = 'dataset'
 
     id = Column(GUID, primary_key=True, index=True, default=uuid.uuid4)
+    person_types = Column(ARRAY(String(255)))
```
seerocode: Looks like using an array is a problem so will need to refactor this to get the tests to pass. Doing a new table is a pain but might be better in the long run.

api/schema.graphql
```diff
@@ -225,6 +227,7 @@ input EntryInput {
   id: ID
   categoryValueId: ID!
   count: Int!
+  personType: String
```
seerocode: I added personType to the entry so we can update entries based on the type of person they are entering a count for.

seerocode: Adding it to a record would mean users would have to click in and click out to add entries for the same date but different person type.

client/src/pages/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -61,20 +62,55 @@ export const renderForm = (
...
+  if (
+    Array.isArray(metadata?.dataset.personTypes) &&
+    metadata?.dataset.personTypes?.length
+  ) {
+    let startIndex = 0;
+    const addDataEntriesCollectionLen =
+      metadata?.dataset?.personTypes?.length *
+      metadata?.dataset?.program?.targets.length;
+
+    while (startIndex < addDataEntriesCollectionLen) {
```
seerocode: There's probably a better way to set this index (where the number of items for "adding" entries is going to be the number of targets multiplied by the number of person types).

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+          aria-labelledby={item.categoryValueLabel}
+        >
+          *
+        </span>
+      </label>
+    ));
+  };
+
+  // returns person type groups if personTypes array
+  if (Array.isArray(personTypes) && !personTypes?.length) {
```
seerocode: There's some duplication here between the return functions in this component that could be cleaned up. Noting it here in case I'm overlooking something obvious that could quickly clean this up.

api/database.py
seerocode: Created a table instead.

api/database.py
```diff
@@ -74,6 +74,20 @@
...
                      )
 
+dataset_person_types = Table('dataset_person_type', Base.metadata,
+                            Column('dataset_id', GUID, ForeignKey(
+                                'dataset.id'), index=True),
+                            Column('person_type_id', GUID, ForeignKey(
+                            'person_type.id'), index=True),
+                        )
+
+entry_person_types = Table('entry_person_type', Base.metadata,
```
jnu: i'm curious why this is many:many? shouldn't it just be one person_type associated with one entry? no need for a mapping table for that, just have a foreign key column on the Entry referencing the PersonType table.

api/database.py
```diff
@@ -459,7 +474,8 @@ class Entry(Base, PermissionsMixin):
     record_id = Column(GUID, ForeignKey('record.id', ondelete="cascade"), index=True)
     inputter = relationship('User')
     inputter_id = Column(GUID, ForeignKey('user.id'), index=True)
-    person_type = Column(String(255), nullable=True)
+    person_types = relationship('PersonType', back_populates='entries')
```
jnu: ah, i see the mapping table is not being referenced here. i would delete the mapping table above and call this `person_type` singular, since a single entry can only have one associated person type.

the relationship is `many:1` -- so many entries will be associated with one single PersonType row. I would expect data access to typically be in that direction (i.e., joining the PersonType to the Entry table), and the reverse relationship of querying all the entries associated with a singular PersonType may never happen in practice, since for reporting we would probably always want several other filters (e.g. by program or region or tag or whatever), bu it's fine to keep `PersonType.entries` if you want.

jnu (changes requested):

api/database.py
```diff
@@ -29,7 +29,7 @@
 from sqlalchemy import create_engine, event
 from sqlalchemy.sql import func
 from sqlalchemy.sql.expression import text
-from sqlalchemy.sql.sqltypes import TIMESTAMP
+from sqlalchemy.sql.sqltypes import ARRAY, TIMESTAMP
```
jnu: can you clean up the ARRAY type here?

api/schema.graphql
```diff
@@ -115,6 +115,14 @@ type Dataset @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   records: [Record!]!
   tags: [Tag!]!
   sumOfCategoryValueCounts: [SumEntriesByCategoryValue!]!
+  personTypes: [PersonType!]
```
jnu: this should be `[PersonType!]!` i think since it'll never be null, right?

api/schema.graphql
```diff
@@ -135,6 +143,7 @@ type Entry @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   categoryValue: CategoryValue!
   count: Int!
   inputter: User!
+  personType: PersonType
```
jnu, line 150: should this be `PersonType!` since it'll never be null?

client/public/locales/en-gb/translation.json
```diff
@@ -2,5 +2,17 @@
   "program": "programme",
   "program_plural": "programmes",
   "Program": "Programme",
-  "Program_plural": "Programmes"
+  "Program_plural": "Programmes",
+  "aboutAttribute": "About {{attribute}}",
+  "attributeDescription": "{{description}}",
+  "cancelAndReturnToDashBoard": "Cancel and Return To Dashboard",
```
jnu: it looks like these all duplicate the `en/translation.json`? we only have to specify translations here if they differ from the general english translation.

client/src/components/PageTitleBar/PageTitleBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+const { Text } = Typography;
+
+// interface inherits the properties of PageHeaderProps
+// and requires the originally optional title and subtitle properties
+interface DefaultProps extends Omit<PageHeaderProps, "title" & "subTitle"> {
+  title: React.ReactNode;
+  subtitle: React.ReactNode;
+}
+
+export const PageTitleBar: React.FC<DefaultProps> = (
```
jnu: it looks like this component just provides styling. what's the advantage of writing a new component for this vs. simply overrriding the `.ant-page-header` class style in css?

client/src/components/PageTitleBar/PageTitleBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  titleBarProps: DefaultProps
+): JSX.Element => {
+  return (
+    <PageHeader
+      {...titleBarProps}
+      className="pageTitleBar"
+      title={<Text style={{ fontSize: "x-large" }}>{titleBarProps.title}</Text>}
```
jnu: I'm not sure what the PageHeader component renders internally, but for accessibility it's better to use semantic header rank tags like `<h1 />` instead of simple Text elements (i assume antd just uses <span> for these).

https://www.w3.org/WAI/tutorials/page-structure/headings/

client/src/pages/DataEntry/DataEntryAggregateDataEntryForm.tsx
jnu: hm i'm wondering why the startIndex is needed at all? isn't that just the same as the array index of the object? either way i don't see what wrapping in the `while` loop adds, since every single time this code runs the check will be false on the first pass and true on the second pass.

simpler:

```ts
for (const personType of metadata?.dataset.personTypes) {
  for (const target of metadata?.dataset.program.targets) {
    form.push({
      /* type + target + count stuff. i wouldn't include the index. */
    });
  }
}
```
then if you need the index later you can get it when you iterate:
```ts
form.map((formItem, index) => /* stuff */);
```
 

but maybe i'm missing something? if you do the need the index, you could also just set `let index = 0` before the loop and set `index: index++` as you go, no need for the outer while loop.

client/src/pages/DataEntry/DataEntryAggregateDataEntryForm.tsx
```diff
@@ -61,17 +62,51 @@ export const renderForm = (
...
-    form = metadata?.dataset?.program?.targets.map((target, index) => ({
-      index: index,
-      category: target.categoryValue.category.name,
-      description: target.categoryValue.category.description,
-      categoryValueId: target.categoryValue.id,
-      categoryValue: target.categoryValue.name,
-      categoryValueLabel: target.categoryValue.name.replace(/\s+/g, "-"),
-      count: 0,
-    })) as Array<Entry>;
+    metadata?.dataset?.program?.targets.map((target, index) => {
```
jnu, line 94: i'm still a little confused why personType would be null - but if so, you could combine this with the upper loop by doing:

```ts
const personTypes  = metadata?.dataset.personTypes || [undefined];
for (const personType of personTypes) {
  for (const target of metadata?.dataset.program.targets) {
    form.push({
      // target stuff
      count: 0,
      personType, // this will be `undefined` when no personTypes are specified, and therefore semantically the same as what you have here.
    });
  }
}
```
it's nice eliminate conditional code paths like this when possible so there's less code to keep in sync.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+
+  const inputs = (entries: any) => {
+    return entries.values.map((item: any, index: number) => (
+      <label
+        key={item.index}
+        id={item.categoryValueLabel}
+        htmlFor={item.categoryValue}
+        className="data-entry-form_label"
+      >
+        <input
```
jnu: is it possible to use the antd `Input` component here to keep styling consistent?

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+          type="text"
+          inputMode="numeric"
+          pattern="[0-9]*"
+          placeholder="0"
+          value={item.count}
+          onChange={(e) => onValueChange(e, item.index)}
+        />
+        {` ${item.categoryValue} `}
+        <span
+          className="data-entry-form_required-field"
```
jnu, line 110: fwiw antd Forms provide validation to address the required field use case

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
   onValueChange,
+  personTypes,
 }: DataEntryCategorySectionProps): JSX.Element => {
   const { t } = useTranslation();
 
   const categories = groupByCategory(entries);
 
+  const categoriesAndPersonTypes = groupByCategoryAndPersonType(entries);
+
+  const inputs = (entries: any) => {
```
jnu, line 68: do we have a more specific type for this?

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+    return entries.values.map((item: any, index: number) => (
+      <label
+        key={item.index}
```
jnu: is this the only place we use the `item.index` i commented on above? in this case it's actually better to not use the numeric index anyway.

react requires a unique key of some sort, the more semantically meaningful you can make those keys the better react will perform. using numeric index keys is a last resort because they index is arbitrary -- the list could be reordered without affecting the elements, which thrashes performance and can also lead to incorrect rendering. (more info here: https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318).

i think for these you want the key to be some combination for `personType + target`, right? the personType name and the target categoryValueLabel would ideally be the key.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+        id={item.categoryValueLabel}
```
jnu: html IDs should be unique across the entire page, which is not true here when there are multiple personTypes. this ID should probably just be the same as the key you use

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+        htmlFor={item.categoryValue}
+        className="data-entry-form_label"
+      >
+        <input
+          key={index}
```
jnu: fwiw you only need keys when you render a list of components. (i.e., only the outermost component in each list item needs a key.)

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+          name={item.categoryValue}
+          required
+          aria-labelledby={item.categoryValueLabel}
+          aria-required="true"
+          type="text"
+          inputMode="numeric"
```
jnu: antd has a NumberInput component, does that work here?

the numeric input component used to vary pretty significantly between browsers. maybe that's fixed, my gut is that it'd be better to use a component from the framework for visual consistency + cross browser compatibility.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -15,73 +16,154 @@ const groupByCategory = (entries: Entry[]) => {
...
+          pattern="[0-9]*"
+          placeholder="0"
+          value={item.count}
+          onChange={(e) => onValueChange(e, item.index)}
```
jnu, line 106: `item.index` will be the same as `index` right?

client/src/pages/DataEntry/DataEntryCategorySections.tsx
jnu: hm, when is this condition true? "personTypes is an array and has no elements in it" -- i thought either personTypes was `null` if the dataset didn't use personTypes, or it'd have some elements in it?

jnu: in general i would think it's possible to reuse the code by checking if personTypes is missing, and if so, add a trivial generic personTypes list of length 1 (as i mentioned above with using `[undefined]` as the base case to reuse code in the loops).

jnu (changes requested):
looks pretty close, great work! i made a few comments to hopefully help clean it up a bit, let me know if there's any way i can help!

api/database.py
seerocode: This was already deleted in recent commits.

api/schema.graphql
seerocode: It can be null if no person Type is assigned to a dataset.

api/schema.graphql
seerocode: It can be null if no person type exists for a dataset so it will not exist for the entry.

client/src/components/PageTitleBar/PageTitleBar.tsx
seerocode: Because this component is used in other places too (like Admin side) and styling is inconsistent. You'd have to copy the style code over each time you use the page header. This also requires the title and subtitle instead of making it optional again for consistency.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: Yeah this was carryover while I was moving things around.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: I'm hesitant to use it because I've seen accessibility issues with it. Namely that scrolling up or down on a mouse inadvertently moves numbers up and down possibly without the user realizing. If this is disable by default in the component then I can try it.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: I thought so but there's a glitch with indices loading out of order so this makes it consistent.

api/schema.graphql
jnu: the way many:many relationships work in sqlalchemy, `Dataset.person_types` will just be an empty list if there's nothing assigned (not `None`). since this is reliably the case, if you add the `!` type restriction on the list you won't have to make `null` checks everywhere in typescript when you want to reference this list. (it'll also keep us from trying to give a `null` value any semantics, since there will never be a null value for this field.)

client/src/components/PageTitleBar/PageTitleBar.tsx
jnu: > you'd have to copy the style code over each time you use the page header.

what do you mean by that? css styles are global, you could define a style for `.ant-page-header` in any css file and it'll apply to everywhere that class is used in the application.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
jnu: this sound potentially like a bug that can happen when you use the list index as the `key` prop -- maybe try changing that and see if it's still broken?

client/src/pages/DataEntry/DataEntryCategorySections.tsx
jnu: i don't have a mouse wheel to test with (i'm not sure if trackpads behave differently), but it sounds like that's the behavior of `<input type="number" />` in some browsers too? https://stackoverflow.com/questions/9712295/disable-scrolling-on-input-type-number

either way it should be easy to disable with `onScroll={e => e.preventDefault()}` or something to that effect.

also again i'd just be cautious that the native `input type="number"` is not consistently implemented across browsers either, so there are potentially accessibility and usability concerns with that route too. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs

client/src/pages/DataEntry/DataEntryAggregateDataEntryForm.tsx
seerocode: Makes more sense. Thanks! I've been looking at this for too long.

seerocode: I also wanted to avoid the nested loop but it's more readable.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: the schema returns an empty array if no person types have been defined for the dataset:

image

client/src/components/PageTitleBar/PageTitleBar.tsx
seerocode: I see what you mean. I can make that global change then although I preferred to isolate it to a component so there's no confusion if someone wants to use the page header in another component that doesn't require that level of consistent styling. Although because CSS cascades, I suppose that won't be an issue.

Here's an example of how it looks now btw:

User -
image

Admin - 
image

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: I added the rule and left this asterisk label for visual explicitness that this field is required. I prefer the labels on the right how we decided in our design so it could read as "X people" instead of "people X".

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: Thank you. I noticed this too while debugging the tests.

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: Yeah, exactly. I picked up the same while debugging tests.

seerocode:
Not sure what's up with the Python tests! @nmanduj1 @jnu any ideas?

api/schema.graphql
```diff
@@ -119,6 +119,14 @@ type Dataset @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   records: [Record!]!
   tags: [Tag!]!
   sumOfCategoryValueCounts: [SumEntriesByCategoryValue!]!
+  personTypes: [PersonType!]!
+}
+
+type PersonType {
+  id: ID!
+  dataset: Dataset!
+  personTypeName: String!
```
jnu, line 128: this could just be `name`, `personType` is redundant with the name of the type

jnu (approved):
thanks ro! 

it does look like all my lingering comments are pretty minor so we should go ahead and merge! awesome work!

client/src/components/PageTitleBar/PageTitleBar.tsx
seerocode: This was updated btw!

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: It still didn't work without the explicit index so I tabled it to get it through for now. Added a story for it: https://app.clubhouse.io/stanford-computational-policy-lab/story/337/indices-for-entries-in-data-entry-form-load-out-of-order

client/src/pages/DataEntry/DataEntryCategorySections.tsx
seerocode: This was changed to ant design input as well.

PR #86, opened 2021-07-07
Admin UI: manage program(mes)

jnu:
UI for managing programs, with some API changes as well to support this.

In this PR:
 - [x] Admin page to list all programs
 - [x] GraphQL query to fetch all programs
 - [x] GraphQL query to fetch single program
 - [x] Admin form to create a new program
 - [x] Admin form to edit an existing program
 - [x] Admin button to delete an existing program
 - [x] Admin button to restore a deleted program
 - [x] Tests for the graphql schema additions
 - [x] UI tests for the new admin pages


client/src/pages/Admin/EditUser.tsx
```diff
@@ -283,7 +284,7 @@ export const EditUser = () => {
   return (
     <div className="admin user-edituser_container">
       <PageHeader
-        onBack={() => window.history.back()}
+        onBack={() => history.push("/admin/users")}
```
nmanduj1, line 287: Why use history vs window.history?

api/database.py
```diff
@@ -246,7 +250,7 @@ class Target(Base, PermissionsMixin):
     id = Column(GUID, primary_key=True, default=uuid.uuid4)
     program_id = Column(GUID, ForeignKey('program.id'), index=True)
     program = relationship('Program', back_populates='targets')
-    target_date = Column(DateTime, nullable=False)
+    target_date = Column(DateTime, nullable=False, server_default=func.now())
```
seerocode: Hm, what's the purpose of setting the default to be current datetime instead of requesting that the user set a target date? I also, see a case where users might not want to set a target date so it would be like "for all time" unless we want to force them to choose some date in the future.

client/src/router/Router.tsx
```diff
@@ -113,11 +118,13 @@ export function RenderRoutes({
   const WrappedPrivateAdmin = (props: AnyRouteProps) =>
     auth.isLoggedIn() ? (
       <Container>
-        {auth.isAdmin() ? (
-          <ProtectedRoutes routes={adminRoutes} />
-        ) : (
-          <div>{t("notAuthorized")}</div>
-        )}
+        <ErrorBoundary>
```
nmanduj1, line 121: General q: what's ErrorBoundary doing here? // How does it work?

api/mutations.py
```diff
@@ -296,55 +296,99 @@ def resolve_delete_team(obj, info, id):
...
-        setattr(program, param, input[param])
-    session.add(program)
+    program = session.query(Program).get(input.pop('id'))
+    
+    if 'team_id' in input:
+        program.team = session.merge(Team(id=input.pop('team_id')))
+
+    if 'targets' in input:
+        program.targets = []
+        # TODO: We want to keep a record of all changes to targets so we can
```
seerocode, line 348: I _think_ we have a story for this? If not, should definitely add and link in this TODO if it can't be covered in this PR.

nmanduj1 (approved):
🚢 nice

api/schema.graphql
```diff
@@ -72,7 +72,8 @@ type Program @needsPermission(permission: [ADMIN, TEAM_MEMBER]) {
   team: Team!
   datasets: [Dataset!]!
   targets: [Target!]!
-  tags: [Tag!]
+  tags: [Tag!]!
+  deleted: DateTime
```
seerocode, line 76: Are we using this field on the UI anywhere?

client/public/locales/en-gb/translation.json
```diff
@@ -12,5 +12,31 @@
   "editData": "Edit",
   "backToDashboard": "Back To Dashboard",
   "noDataAvailable": "No Data Available",
-  "adminSidebarManagePrograms": "Programmes"
+  "adminSidebarManagePrograms": "Programmes",
+  "admin": {
+    "program": {
+      "index": {
+        "title": "Programmes",
```
seerocode: There's a lot of duplication here with the word "programmes". I think using a "programmes" field in this json should be sufficient if we use string literals to form the rest of the required string on the UI.

seerocode: We can also use params here so you can do something like `t({"Edit a"} "programme")`. I can't remember the exact syntax but it's something like that.

client/src/__tests__/AdminProgramList.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+import { tick } from "./utils";
+
+import { ProgramList } from "../pages/Admin/ProgramList";
+import { ADMIN_GET_ALL_PROGRAMS } from "../graphql/__queries__/AdminGetAllPrograms.gql";
+import { ADMIN_GET_ALL_TEAMS } from "../graphql/__queries__/AdminGetAllTeams.gql";
+import { ADMIN_CREATE_PROGRAM } from "../graphql/__mutations__/AdminCreateProgram.gql";
+import { ADMIN_GET_PROGRAM } from "../graphql/__queries__/AdminGetProgram.gql";
+
+const TEAMS = {
```
seerocode, line 16: You can also use the custom mock provider here and update the mocks to avoid adding a lot of test data to the tests! Not necessary to change here and pointing it out for the future.

client/src/components/Error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export type ErrorBoundaryProps = {
+  children: React.ReactNode;
+};
+
+/**
+ * Class to handle errors thrown during rendering.
+ */
+export class ErrorBoundary extends React.Component<
```
seerocode, line 15: Because of the class component this should catch async errors as well?

client/src/graphql/hooks/useQueryWithErrorHandling.ts
```diff
@@ -0,0 +1,56 @@
...
+    console.warn(e);
+  });
+
+  return result;
+};
+
+/**
+ * Variation of apollo's useQuery hook with automatic error handling.
+ *
+ * Errors get thrown, presumably will be caught by an ErrorBoundary in the
```
seerocode, line 43: Can we add a test to confirm this?

client/src/pages/Admin/CreateProgram.tsx
```diff
@@ -0,0 +1,234 @@
...
+   * Form object that will be filled out by the UI.
+   */
+  form: FormInstance<CreateProgramFormValues>;
+};
+
+export const CreateProgram = ({ form }: CreateProgramProps) => {
+  const [saving, setSaving] = useState(false);
+  const [saveError, setSaveError] = useState<Error | null>(null);
+  const { t } = useTranslation();
+  const apolloClient = useApolloClient();
```
seerocode, line 53: I haven't seen this before! What's this for vs the general query/mutation run?

client/src/pages/Admin/EditUser.tsx
seerocode, line 287: This might be confusing for the user if they're taken back to a page that wasn't the page they were on. In case they can click into this edit user window from a different origin.

seerocode: The breadcrumbs should allow them to go back to the users page just as well!

jnu: the [`history`](https://github.com/ReactTraining/history) library is better because it works in all environments (including tests). i also changed the behavior a bit to not be a "back" button but instead just navigate to the index page

client/src/router/Router.tsx
jnu: you can think of it like a `try...catch` block for the react render cycle. if any descendant component under the ErrorBoundary throws an error during render it can be handled by this ErrorBoundary component and displayed as an error message (and logged to the Sentry, or whatever).

this simplifies error handling in the children components -- you can just `throw` whatever issues come up and not worry about showing an error message in that specific component, e.g.:

```ts
const MyCmp = () => {
    const { data, error, loading } = useQuery(MyGqlQuery);
    if (error) {
        // just throw the error and let ErrorBoundary handle it, don't include any
        // specific error UI here.
        throw error;
    }
   return (
        <Container loading={loading}>
            <Data data={data} />
        </Container>
    );
};
```

api/database.py
jnu: ah, i probably misunderstood this column. it means "date we want to reach this target by?"

this would require a new UI input and we'd have to figure out what to do when the date goes by (do we want to prompt people to set new targets, or what?)

i'll change this to be nullable with no default and put a ticket to discuss as a possible future enhancement.

api/mutations.py
jnu: i made a story for it the other day: https://app.clubhouse.io/stanford-computational-policy-lab/story/324/keep-historical-record-of-targets-in-the-database

i'll link it here

api/schema.graphql
jnu: yep, it's used to determine whether to show the "inactive" view (disabled form fields and an alert that prompts user to restore the program if they need to)

api/database.py
seerocode: That's how I saw the target too. If the target goes by we can have some kind of indicator that says "out of date" or "within target date" and let the user figure out what to do with that information afterward. The data will keep coming in no matter if the target date has passed or not.

client/public/locales/en-gb/translation.json
jnu: ah found it, https://www.i18next.com/translation-function/nesting

```ts
t("Edit a $t(program)")
```
good idea, i will update

client/src/router/Router.tsx
nmanduj1: interesting, will it also catch an error within itself? // is it worth testing that this behaves the way we're expecting? // is there a way to test this?

client/src/components/Error/ErrorBoundary.tsx
jnu: no, error boundaries can't catch async errors (you can only catch async errors if you have an async function -- `render` is still synchronous). the class is required to implement the ErrorBoundary lifecycle methods - more info here: 
https://reactjs.org/docs/error-boundaries.html

jnu: fwiw a common pattern for async errors is to store the error in state. so you could write a hook like:

```ts
const useHook = () => {
  const [data, setData] = useState(null);
  const [loading, setLoading] = useState(false);
  const [error, setError] = useState(null);
 
  setLoading(true);
  myAsyncFunction()
    .then(d => setData(d))
    .catch(e => setError(e))
    .then(() => setLoading(false));
  return {data, loading, error};
};
```
then in your component if you wanted to use the ErrorBoundary to show the error you could do it by throwing the error synchronously when the state value gets set:

```ts
const MyCmp = () => {
  const {data, loading, error} = useMyHook();
  if (error) {
    throw error;
  }

  return <Container loading={loading} data={data} />;
}

client/src/graphql/hooks/useQueryWithErrorHandling.ts
jnu: maybe i could rephrase the comment, the point is just to say that errors are not handled directly within this component and it should be used with an ErrorBoundary (this is basically the same advice react prints out in the console whenever it sees an unhandled error). if we wanted to test we're using an ErrorBoundary and it's handling errors, we'd have to write an integration test.

client/src/pages/Admin/EditUser.tsx
jnu: i found the strict "back" behavior pretty unintuitive in every situation except for the one where "back = the index page." e.g. if you copy+paste the link into the browser, "back" takes you to whatever you were looking at before (not relevant). another very common thing is to be taken to the login screen (since the login screen redirects you to whichever page you were trying to access), which was also useless. 

i do agree it'll be redundant with breadcrumbs, but i temporarily had to take breadcrumbs out of the admin page because there's a bug (third-level nested routes were not getting rendered at all).

client/src/graphql/hooks/useQueryWithErrorHandling.ts
seerocode: Right, we'd have to simulate an error somehow for the actual error boundary. Was it working if you manually tested it? I'm assuming it came up in dev and having the error boundary was helpful to catch stack traces. 

I had saved this video for eventual implementation of an error boundary! https://www.youtube.com/watch?v=dxWrHEOD5DU

client/src/components/Error/ErrorBoundary.tsx
seerocode: Ah, ok. I have a similar error state in the data entry component for this reason but using try catches.

client/src/pages/Admin/EditUser.tsx
seerocode: OK. I liked the back button in the header component AntD had but for similar reasons, I removed it in favor of breadcrumbs. I added a story to fix the breadcrumbs in the admin UI when I saw they weren't working so this is good context to understand why.

client/src/pages/Admin/CreateProgram.tsx
jnu: the useQuery and useMutation hooks are convenient for the common flow when you just have a component that just fetches or saves data and renders the result.

in this component we need more control over running queries -- to implement the "template" feature we need to run a query to fetch the program that was specified as a template in the UI (which obviously we don't know in advance), derive some info from it, then run the mutation. we also want to treat that query+mutation as a unit (handle errors/success as if it were one operation).

you might be able to do this with hooks too, i just thought it was easier/clearer to use the client directly.

api/database.py
jnu: i like that this tries to motivate people to keep updating their goals.

a couple practical things we'll have to figure out off the top of my head:
 - does the target date really apply to a single target, or to a set of them? (in the db the column is currently on the target table, so each could theoretically be set individually, but in practice targets don't exist in isolation, they're zero-sum: you can't increase one without decreasing another). in the ui maybe it makes sense to treat the target date as applying to the entire program, not just a single target row.
 - can you edit targets before the target date? does this result in a new Target row being created (like we're currently planning to do for edits), or does it update the existing row?
 - how do we surface alerts about expiring targets so that people can take action and reset them? if the alerts are buried on the edit program page they are likely to go unnoticed; but on the flip side we also have to worry about alert fatigue.

i'd love to discuss more, i'll put it on the backlog for now

seerocode (approved):

client/src/router/Router.tsx
jnu: you mean does it catch errors within the ErrorBoundary implementation? no it doesn't, but this is true of any ordinary try..catch block too (if an error happens within a catch block it will not "catch" itself).

you can definitely write a unit test for the component ... im not very concerned about a dedicated test right now (versus the app snapshot test where the basic rendering is implicitly covered), because the implementation is trivial (it's even simpler than the example given in the [react docs](https://reactjs.org/docs/error-boundaries.html)). if we were to add actual server logging it could be worth adding a test.

client/src/graphql/hooks/useQueryWithErrorHandling.ts
jnu: yea, i tested manually.

i'd think of the error boundary as similar to a try-catch block. we don't need to test that the error boundary catches errors, we can trust that it does (it's part of the react framework). but in the same way with try-catch, we _do_ want to make sure we're using the error boundary at the right place in the application, but that's really an integration test and not a unit test.

PR #89, opened 2021-07-14
Ch309/nginx routing

nmanduj1:


client/package.json
```diff
@@ -34,7 +34,7 @@
     "start": "react-scripts start",
     "start-env": "sh -ac '. .env.${REACT_APP_ENV}; react-scripts start'",
     "start:mock": "REACT_APP_ENV=mock yarn run start-env",
-    "build": "react-scripts build",
+    "build": "GENERATE_SOURCEMAP=false react-scripts build",
```
jnu: why is this necessary to turn off? sourcemaps are very helpful!

nginx.conf
```diff
@@ -44,25 +44,15 @@ http {
             add_header Cache-Control max-age=31536000;
         }
 
-        location /graphql/ {
-            # Proxy GraphQL requests to API server
-            proxy_set_header X-Forwarded-For $remote_addr;
-            proxy_set_header Host $http_host;
-            proxy_pass http://api$uri;
+        location = /api {
+            return 302 /api/;
```
jnu: i think technically `301` is better since this redirect will be permanent, but not a big deal

jnu (approved):
looks awesome! i'm curious why u had to remove the sourcemaps tho -- would be great to keep those if possible, they make debugging many many times easier!

client/package.json
seerocode: Is this related in any way to the canvas package issue when building locally with docker?

seerocode (approved):

client/package.json
nmanduj1: Generating the source maps was breaking the build when running docker locally. Hm, reflecting on this a bit more- in addition to setting this flag I also increased the memory resource allocation from 2gb to 4 on my computer. I guess it's possible upping that memory resource allocation did the trick and the sourcemaps are a red herring. I will try running this while generating the sourcemaps to confirm it is a sourcemap issue, or not.

Regardless, I figure if we're running the build in a Docker container, chances are we are expecting an optimized production build. In this case we probably do not require the sourcemaps to be compiled. Alternately, we can set the `GENERATE_SOURCEMAP` flag as an ENV directly in our Docker file...

nmanduj1: 🎉 confirmed!! It was a memory allocation issue- Since it sounds like you'd prefer the sourcemaps be generated, I will remove this flag and merge.

nmanduj1: oop, sorry Ro, @seerocode ! missed your comment- it ended up being an issue with memory allocation. Which, I guess may have to do with the canvas package if the package is so big that it's consuming lots of memory to build and run.

jnu: great to hear, thanks for figuring that out!

(and yes anyone who needs to debug this application will appreciate having the sourcemaps -- trying to debug what's wrong with typescript code by looking at the minified javascript is incredibly time consuming and tedious, unless you have sourcemaps!)

seerocode: @nmanduj1 thank you! How did you adjust memory allocation so I can do the same?

nmanduj1: There are a few options, I just went through and upped it for all my images through Docker Desktop- right click on the Docker ship icon in your menu bar to get Preferences then navigate to Resources. 
There's a way to set individual memory specifications in our docker-compose file- we might want to go that route .. although our global setting trumps the individual setting so setting it in that way may be a moot point.

If you continue running into problems at build time, try manually removing both volumes before running `docker-compose up --build` : 
`docker volume rm -f bbc-50-50_pgdata` and `docker volume rm -f bbc-50-50_staticFiles`

PR #90, opened 2021-07-15
[Ch293] Admin - teams management

jnu:
Last major piece of the admin functionality, creating and editing teams.

Included:
 - [x] Teams index page
 - [x] Create new team form
 - [x] Edit team form
 - [x] UI tests

Not included in this PR (to help keep it small, will do in a follow-up):
 - Delete team
 - Restore team


jnu:
image

client/src/__tests__/fixtures.ts
```diff
@@ -0,0 +1,118 @@
+export const TEAMS = {
```
seerocode, line 1: This is to be used for the MockProvider? Seems like we'd be maintaining two different mock fixtures because `MockResolvers` also exists and I don't want to maintain two different sets of mock data for tests.

jnu: it looks like the MockResolvers is trying to create fake graphql server that can respond to any graphql query, is that right?

what i like about the MockedProvider is that you have to state explicitly which queries should run and what responses you should get. so if an unexpected query is running the test will fail, instead of just handling the query and returning an ostensibly valid result. this is implicitly what a lot of these tests rely on, e.g. if the form failed validation it shouldn't send a query. this also lets you simulate different error conditions easily, since you can set different responses for the same query -- can you do that in the MockResolvers too?

agree it'd be good to not have too much test data to update, tho it seems like these serve slightly different purposes.

seerocode: Yeah, it mocks graphql data. I created the wrapper following the example here from [Stripe](https://www.freecodecamp.org/news/a-new-approach-to-mocking-graphql-data-1ef49de3d491/) because the current MockProvider from Apollo was limiting in that it was difficult to mock deleting a record from the records table because you couldn't simulate a different result for the same query after a re-render. 

If this works for a separate use case then it's fine. I want to make sure we aren't duplicating work so it's a cohesive project.

seerocode (approved):

client/src/__tests__/fixtures.ts
jnu: > MockProvider from Apollo was limiting in that it was difficult to mock deleting a record from the records table because you couldn't simulate a different result for the same query after a re-render.

hm, that should not be a problem -- you have to specify the query and the result for every individual time the query is run -- if you refetch data you will need to specify the same query twice or you will get an error about an unexpected query. (this is exactly what i like about it! and it's how i test that the form only gets submitted once in the tests.) wonder if this was a caching problem or maybe an `await` problem?

PR #95, opened 2021-07-16
Prompt when leaving page with unsaved changes

jnu:
Shows a confirm dialog to leave a page when there are unsaved changes.

Also added some visual consistency to the three different pages with other confirmation modals and button disabled state.

image


nmanduj1 (approved):
Look fine to me - 🚢

PR #96, opened 2021-07-20
Autocomplete for category values

jnu:
1) Adds autocomplete in UI for suggesting CategoryValues when setting targets, so it's easier to set up programs
2) Validate that category values are not added multiple times as targets
3) Ensure that category values get deduped on the backend when inserting into the database
4) Fix bugs with normalizing names (need to trim space before capitalizing, or capitalize won't do anything)
5) Add unique constraint on (category, category_value_name)

### DIFF
--- a/api/database.py
+++ b/api/database.py
@@ -229,7 +229,7 @@ class Tag(Base):
     __tablename__ = 'tag'
     
     @classmethod
-    def get_or_create_tag(cls, session, tag_dict) -> "Tag":
+    def get_or_create(cls, session, tag_dict) -> "Tag":
         '''Upsert partially-specified tags.
 
         :param session: Database session
@@ -258,7 +258,7 @@ class Tag(Base):
     
     @classmethod
     def clean_name(cls, name):
-        return name.capitalize().strip()
+        return name.strip().capitalize()
 
     @validates('name')
     def capitalize_tag_name(self, key, name):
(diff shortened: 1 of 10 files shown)


api/database.py
```diff
@@ -361,7 +364,53 @@ class CategoryValue(Base):
     
     @validates('name')
     def capitalize_name(self, key, name):
-        return name.capitalize().strip()
+        return self.clean_name(name)
+
+    @classmethod
+    def clean_name(cls, name):
```
nmanduj1, line 370: Shouldn't this be an instance method? What's the use case for passing the name back into the class to clean it?

nmanduj1 (approved):
🚢 - Nice optimizations ! 🎉

api/database.py
```diff
@@ -361,7 +364,53 @@ class CategoryValue(Base):
...
+
+    @classmethod
+    def get_or_create(cls, session, spec):
+        """Idempotently create the category value.
+
+        :session: Database session
+        :spec: Dict containing category value parameters
+        :returns: CategoryValue object
+        """
+        if 'id' in spec:
```
nmanduj1, line 386: spec?

api/database.py
jnu: what do you mean by pass back into the class? (classmethods are effectively just pure functions that are grouped with the class for logical reasons.)

a few reasons to write it as a classmethod:
1) you can't reference instance methods from other classmethods, so if you want to write something like `CategoryValue.get_by_name` (which obviously can't be an instance method) you need to have this as a classmethod. (we need `get_by_name` in order to fix the bug where we could accidentally enter duplicate CategoryValues.)
2) the `name` attribute of the instance (which represents the database column) already uses a validator (which uses this clean_name function) to keep the name properly formatted
3) even in oop pure functions are simpler to use and test than instance methods, since their output is completely determined by their input, not on the state of the instance they're attached to

api/database.py
jnu: short for "specification" ... it's a variable name you'll see sometimes for a "a set of values that will be used to specify an instance" (in place of `data` or `d` or `cv_dict` or other kind of opaque names. it's described on l383)

PR #97, opened 2021-07-20
[ch313] Password strength estimation and validation

jnu:
Add a password strength meter and enforce passwords are complex enough.


client/package.json
```diff
@@ -29,10 +30,11 @@
     "react-scripts": "4.0.1",
     "typescript": "^4.2.4",
     "use-react-router-breadcrumbs": "^2.0.0",
-    "web-vitals": "^0.2.4"
+    "web-vitals": "^0.2.4",
+    "zxcvbn": "^4.4.2"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "react-scripts --expose-gc --max-old-space-size=8192 start",
```
nmanduj1: Were you running into memory issues? // I may have missed it but I don't see you requesting garbage collection to occur- were you using it to debug the memory issues? We might just want to leave that `--expose-gc` flag out if we're not requesting gc.

client/src/pages/ResetAccountPassword/PasswordStrength.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const usePasswordStrength = () => {
+  const [analysis, setAnalysis] =
+    useState<zxcvbn.ZXCVBNResult | undefined>(undefined);
+  return {
+    analyze: (password: string) => {
+      const result = zxcvbn(password);
+      setAnalysis(result);
+      return result;
+    },
+    analysis,
```
nmanduj1, line 18: zxcvbn is a cool little library!

nmanduj1:
General q: The BBC will integrate with their own authentication system so they will be checking the pw strength and validation on their end- Will this work have to be removed when they integrate the tool with their SSO system?

nmanduj1 (approved):
🚀 No more weak passwords!

client/package.json
jnu: i used this to debug issues from transforming zxcvbn, which uses large dictionaries to inform the strength estimates

jnu:
> General q: The BBC will integrate with their own authentication system so they will be checking the pw strength and validation on their end- Will this work have to be removed when they integrate the tool with their SSO system?

 - if someone forks this and uses sso instead of our login then they won't end up using this
 - id like to leave our account system intact in the main branch and hopefully also support sso in a more general way in the future, so that we can account for as many different deployment scenarios as possible

PR #98, opened 2021-07-21
[ch332] Prompt user to change password on first login

jnu:
- Add `last_login` column to database and track this when user logs in
 - Add `last_changed_password` column to database and track this when user changes password
 - API route to get a password reset token for one's own account if they are logged in
 - Detect when user has never changed password and immediately prompt them to do so when they log in

Not a hugely important feature, but I wanted to get this PR in soon bc it adds database columns.

### DIFF
--- a/api/user.py
+++ b/api/user.py
@@ -143,9 +143,9 @@ class SQLAlchemyORMUserDatabase(BaseUserDatabase):
         # Only allow updates of certain columns
         for k in ['first_name', 'last_name', 'email', 'is_active', 'is_verified', 'hashed_password']:
             if k in d:
-                setattr(dbuser, k, d[k])
-                if k == 'hashed_password':
+                if k == 'hashed_password' and dbuser.hashed_password != d[k]:
                     dbuser.last_changed_password = func.now()
+                setattr(dbuser, k, d[k])
 
                 # Special handling for deletes, since we use the `deleted`
                 # column while the fastapi-users library uses is_active. We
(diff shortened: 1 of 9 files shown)


api/user.py
```diff
@@ -174,5 +181,19 @@ async def delete(self, user: UserModel) -> None:
...
+    async def authenticate(self, credentials) -> Optional[UserDBModel]:
+        user = await super().authenticate(credentials)
+        # Mark this as the most recent login.
+        # The user object we return will have the last login, not this one.
+        session = self.session_factory()
+        session.query(database.User).filter(
+            database.User.id == user.id
+            ).update({
+                "last_login": func.now(),
+                }, synchronize_session=False)
```
nmanduj1, line 193: I believed you advocated for using "fetch" here instead of "False" the last time we spoke about this.

nmanduj1 (approved):
🚢 🚢 🚢  All aboard!

api/user.py
jnu: the right setting always depends on the problem you're trying to solve. in this case, the field describes the last time a user logged in. when a user logs in, we want to update that column, but we _don't_ want to update the instance. (if we did, `last_login` would just tell us the _current_ login timestamp, not the _last_ login timestamp.)

nmanduj1: hm, sounds like maybe we should add a field in the db to log current login and last login to not have to depend on the instance not being synced after logging in. Ive seen this pattern used in many projects including when using Devise on Rails.

jnu: if you want to add something like a user audit table to log all of the sign-in / sign-out actions that'd be great!

PR #99, opened 2021-07-21
[Ch333] UI for bootstrapping on first app load

jnu:
Currently the app can't be used without running our `python database.py --dummy-data` init script. This PR fixes that by adding a simple bootstrapping UI.

It works like this:
1) When the client makes a request to the API, the API determines if it's in the "blank slate" condition by checking if there's anything in the database (currently just looks at the `organization` table -- if that's empty, nothing has been setup). [Also, if the database does not exist at this point it will now be created automatically, with the correct tables. Things like default roles and categories will also be created here.]
2) If the app is a blank slate, the API will throw a `418` error. The client handles this by displaying the bootstrapping form.
3) The user can enter the organization name, their name, and their email address and press submit.
4) The API creates the Organization and an admin User with the given info, and emails a temporary password to the user
5) User can now log in. The app is no longer in the blank slate mode so permissions are now buttoned up as usual.

image

### DIFF
--- a/api/user.py
+++ b/api/user.py
@@ -196,4 +196,5 @@ class SQLAlchemyORMUserDatabase(BaseUserDatabase):
 
         return user
 
-user_db = SQLAlchemyORMUserDatabase(database.SessionLocal)
+
+user_db = SQLAlchemyORMUserDatabase(database.connection)
--- a/api/user.py
+++ b/api/user.py
@@ -1,10 +1,14 @@
 from pydantic import BaseModel, UUID4, EmailStr
 from datetime import datetime
+from fastapi_users.authentication import CookieAuthentication
+from fastapi_users import FastAPIUsers
 from fastapi_users import models
+from fastapi_users.utils import generate_jwt
 from fastapi_users.db.base import BaseUserDatabase
 from fastapi_users.db.sqlalchemy import GUID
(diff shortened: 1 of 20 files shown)


jnu:
I also cleaned up some of the current database connections stuff, in preparation for further reworking in https://app.clubhouse.io/stanford-computational-policy-lab/story/302/use-async-db-calls

api/user.py
```diff
@@ -196,4 +208,38 @@ async def authenticate(self, credentials) -> Optional[UserDBModel]:
...
-user_db = SQLAlchemyORMUserDatabase(database.SessionLocal)
+
+user_db = SQLAlchemyORMUserDatabase(database.connection)
+
+
+# remove cookie_secure=False later for production
+cookie_authentication = CookieAuthentication(
+        secret=settings.secret,
+        lifetime_seconds=3600,
+        cookie_secure=False,
```
seerocode, line 219: Not part of this PR but is it okay to leave the cookie security as false? I know we have a story for this but thinking for handoff.

seerocode (approved):
I was wondering how we were going to set up a first time user. Looks good to me! 

🎉 tests!

api/user.py
jnu: it's not a big deal, i can add a config option tho

PR #103, opened 2021-07-22
Cache-bust translations files

jnu:
Make sure translations files are loaded fresh when we deploy new versions of the app. The cache-busting URL is set by the timestamp of the build.

### DIFF
--- a/client/src/services/i18next.ts
+++ b/client/src/services/i18next.ts
@@ -21,6 +21,12 @@ i18next
       escapeValue: false, // not needed for react as it escapes by default
     },
     debug: process.env.NODE_ENV === "development",
+    backend: {
+      requestOptions: {
+        cache: "no-cache",
+      },
+      queryStringParams: { t: process.env.REACT_APP_BUILD_TS },
+    },
   });
 
 export default i18next;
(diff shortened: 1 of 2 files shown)


client/src/services/i18next.ts
```diff
@@ -21,6 +21,12 @@ i18next
       escapeValue: false, // not needed for react as it escapes by default
     },
     debug: process.env.NODE_ENV === "development",
+    backend: {
+      requestOptions: {
+        cache: "no-cache",
```
seerocode, line 26: How'd you discover this? Also can a ticket be created for historical reference?

seerocode (approved):

jnu:
https://app.clubhouse.io/stanford-computational-policy-lab/story/338/cache-busting-for-i18n-translations

client/src/services/i18next.ts
jnu: the general options are documented here https://github.com/i18next/i18next-http-backend#backend-options

cache-busting techniques are the same for all static assets. pretty much three options: 1) fingerprint the asset in the filename (best bc it refreshes exactly as often as the contents change, but also requires knowing the fingerprinted name elsewhere which can cause a chain of unnecessary cache invalidations, 2) unique query string url (not necessarily optimal for cache of the file itself, but other files don't need knowledge of the specific fingerprint), 3) headers (ideally everything just works with headers but these are not always consistently implemented by browsers/servers).

PR #104, opened 2021-07-22
[ch336] Admin controls for dataset person types

jnu:
- Adds a UI for the admin to manage person types for a dataset
 - Ensures person types get deduped by name on the backend
 - Show all datasets to admin when they login
 - Clean up some more dummy data stuff

### DIFF
--- a/api/queries.py
+++ b/api/queries.py
@@ -9,6 +9,7 @@ from database import (
         Entry,
         Team,
         Role,
+        PersonType,
         Program,
         Organization,
         Tag,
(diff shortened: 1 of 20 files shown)


api/queries.py
```diff
@@ -232,3 +233,13 @@ def resolve_organizations(obj, info):
     '''
     session = info.context['dbsession']
     return session.query(Organization).filter(Organization.deleted == None).order_by(Organization.name.asc()).all()
+
+
+@query.field("personTypes")
```
seerocode, line 238: Person types worked without this field. Why does it need to be sorted in ascending order?

client/src/App.css
```diff
@@ -8,7 +8,7 @@
 }
 
 .ant-layout {
-  min-height: initial;
+  min-height: max-content;
```
seerocode, line 11: 👍🏽

seerocode (approved):

nmanduj1 (approved):
🚢

api/queries.py
jnu: this is a new top-level query that fetches all of the personTypes used in the app. this is used for autocomplete in the admin "edit program" screen to help them reuse existing personTypes.

ascending order for two reasons: 1) ensures determinism in tests, and 2) they will probably always be used alphabetically anyway (e.g. in the autocomplete dropdown)

client/src/App.css
jnu: looked like there was an optimization (or possibly just a bug) in the minified CSS that replaced `min-height: initial` with `min-height: 0`, which breaks styles. this fixes that.

PR #105, opened 2021-07-22
First Valiant Attempt

nmanduj1:


seerocode:
I'm not too familiar with setting up deployment workflows so I'll wait on @jnu also to review.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+          gcloud auth configure-docker
+
+      - name: Push API image to GCR's docker repo
+        if: success()
+        env:
+          GCR_REPO: gcr.io/${{ secrets.GC_PROJECT_ID  }}/bbc5050-api
+        run: |
+          docker tag api:${GITHUB_REF##*/} $GCR_REPO:${GITHUB_REF##*/}
+          docker tag api:${GITHUB_REF##*/} $GCR_REPO:latest
+          docker push $GCR_REPO:${GITHUB_REF##*/}
```
jnu: just curious, what do you expect `GITHUB_REF` to be in this workflow? (in other words, what tag are you trying to create here?)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
+name: Build Workflow
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  deploy:
+    name: Build and Push BBC Reporting Tool Images to GCP
```
jnu: here and elsewhere, i think we should shift away from bbc-specific branding - they have their own repo now, and we are starting to plan for giving this tool to other orgs!

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+          project_id: ${{ secrets.GC_PROJECT_ID  }}
+          service_account_key: ${{ secrets.GCR_ACCESS_KEY }}
+          export_default_credentials: true
+
+      - name: Check out main BBC repo
+        uses: actions/checkout@v2
+        with:
+          ref: ${{github.head_ref}}
+
+      - name: Build main BBC API image
```
jnu: rather than specifying explicitly which containers to build in this file, it would be cool to reuse the `docker-compose.yml` manifest, which already defines these things. two reasons for this:
1) we've talked about changing the containers (e.g., bundle static assets with nginx directly instead of mounting them). it's best if we only have to change one build manifest (the docker-compose.yml) instead of remembering to keep two in sync
2) if the goal is to deploy the app to gcp, then using compose's build command is def the way to go to ensure it's building all the required dependencies.

in order to do this you'll need to add another step (after setting up docker) to make sure `docker-compose` is available on the machine, but then your build steps simplify to just `docker-compose build`. You can look at the `.env` file in the repo to configure image names and tags as necessary (e.g. use the gcr.io repo instead of the public scpl docker hub repo).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
+name: Build Workflow
+
+on:
+  push:
+    branches: [main]
```
jnu: double checking, this will rebuild everything any time we merge code - that means the images we push will often be broken. do we want that, or would we rather run this only when we tag commits (e.g. as `v1.0.2`)?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
+name: Build Workflow
```
jnu: smol q: the file is currently called deploy.yml, could we change it to `build.yml` to reflect what it does (and the name!)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+        if: success()
+        env:
+          GCR_REPO: gcr.io/${{ secrets.GC_PROJECT_ID  }}/bbc5050-api
+        run: |
+          docker tag api:${GITHUB_REF##*/} $GCR_REPO:${GITHUB_REF##*/}
+          docker tag api:${GITHUB_REF##*/} $GCR_REPO:latest
+          docker push $GCR_REPO:${GITHUB_REF##*/}
+          docker push $GCR_REPO:latest
+
+      - name: Push Client image to GCR's docker repo
```
jnu: smol smol wording nit - i would either say "push to GCP's docker repo" or "push to GCR" ...  "GCR's docker repo" doesn't quite make sense in my head (bc GCR doesn't _have_ a repo, it _is_ a repo!) 

(or maybe "our docker repo on GCR" if ur thinking of GCR as a global service, instead of our space within that service ... but im clearly overthinking this :[ )

jnu (commented):
awesome! valiant start indeed! left some comments to help clean this up a bit, lmk if you wanna talk anything through! nice work!

.github/workflows/deploy.yml
nmanduj1: `GITHUB_REF` references the current branch name. Maybe using `GITHUB_SHA` makes more sense...

.github/workflows/deploy.yml
nmanduj1: Hm, this workflow is only going to run when we merge code into main- theoretically we should only be merging code that works/ builds into main. 
Not sure why we should be expecting broken images. 
Can you elaborate?

PR #107, opened 2021-07-23
[ch340] Aggregate person types in table

jnu:
- Sum across personTypes when rendering table
 - Other minor cleanup (getting rid of react render errors, and following up on my comments from #85 )

### DIFF
--- a/client/src/pages/DataEntry/DataEntryCategorySections.tsx
+++ b/client/src/pages/DataEntry/DataEntryCategorySections.tsx
@@ -1,6 +1,6 @@
-import { Card, Col, Row, InputNumber, Form } from "antd";
+import { Card, Col, Form, InputNumber, Row } from "antd";
 import _ from "lodash";
-import { ChangeEvent } from "react";
+import React, { ChangeEvent } from "react";
 import { useTranslation } from "react-i18next";
 import { GetDataset_dataset_personTypes } from "../../graphql/__generated__/GetDataset";
 import { Entry } from "./DataEntryAggregateDataEntryForm";
@@ -11,6 +11,16 @@ interface DataEntryCategorySectionProps {
   personTypes?: ReadonlyArray<GetDataset_dataset_personTypes> | null;
 }
 
+/**
+ * Group of entries related to the given category.
+ */
+interface GroupOfEntries {
+  values: Entry[];
+}
+
+/**
+ * Group entries together by category.
+ */
 const groupByCategory = (entries: Entry[]) => {
   const categoryGroups = _(entries)
     // Group the elements of Array based on nested `category` property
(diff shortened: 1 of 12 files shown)


client/src/pages/DataEntry/DataEntryCategorySections.tsx
```diff
@@ -60,58 +70,31 @@ const DataEntryCategorySections = ({
...
-          item.personType
-            ? `${item.personType.personTypeName} ${item.categoryValueLabel}`
-            : item.categoryValueLabel
-        }
-        htmlFor={
-          item.personType
-            ? `${item.personType.personTypeName} ${item.categoryValueLabel}`
-            : item.categoryValueLabel
-        }
+        key={`${item.personType?.personTypeName}-${item.categoryValueLabel}`}
```
seerocode, line 76: If there's no person type then there would just be a -categoryValueLabel ?

client/src/pages/DatasetDetails/DatasetDetailsRecordsTable.tsx
```diff
@@ -45,14 +45,17 @@ const DatasetDetailsRecordsTable = ({
    * and its count as the corresponding value to create an
    * array of flat objects.
    */
+  console.log("RECORDS", records);
```
seerocode: Ohp, console log left over!

client/src/setupProxy.js
```diff
@@ -4,7 +4,7 @@
  *
  * Based on https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually
  */
-const createProxyMiddleware = require("http-proxy-middleware");
+const { createProxyMiddleware } = require("http-proxy-middleware");
```
seerocode, line 7: This didn't error for you? Maybe I need to pull changes again locally because it yelled at me to remove the curly braces because it wasn't a function.

seerocode (approved):

client/src/pages/DataEntry/DataEntryCategorySections.tsx
jnu: it'll be something like `undefined-Trans men` which isn't the most pretty, but it doesn't matter because it just has to be a string that uniquely identifies the item

client/src/pages/DatasetDetails/DatasetDetailsRecordsTable.tsx
jnu: woops thanks!

client/src/setupProxy.js
jnu: i think this changed in a version of the package, i'd make sure to yarn install again and see if that fixes it

PR #109, opened 2021-07-27
[ch261] Link to datasets filtered by Team name on Homepage when user clicks team name in Sidebar

seerocode:
Story details: https://app.clubhouse.io/stanford-computational-policy-lab/story/261

### Changes:
1. When a user clicks on a team title from the sidebar (e.g. BBC News), the home table will filter by that team name and hide the search bar
image

team name not found:
image

team param value missing:
image

2. Added home button to sidebar to make navigation easier for user
3. When no results are found in search, table shows no data:
image

TODO:
- [X] return no data if filter on home table is not found
- [X] if filtering by team name only, table filter should only search the "team" column


client/src/pages/Home/Home.tsx
```diff
@@ -152,36 +89,56 @@ const Home = (): JSX.Element => {
...
   if (error) return <ErrorFallback error={error} />;
 
   return (
     <>
-      {loading ? (
-        <Loading />
+      {teamNameFilter ? (
+        <Alert
+          style={{ margin: "1rem 0rem" }}
+          message={`Showing datasets for team: ${teamNameFilter.toUpperCase()}`}
```
jnu: can this use the translation file?

client/src/pages/Home/Home.tsx
```diff
@@ -152,36 +89,56 @@ const Home = (): JSX.Element => {
     setFilteredData(filteredData);
   };
 
+  // Searches table for team name if selected from user sidebar
+  const { search } = useLocation();
+  const teamNameFilter = new URLSearchParams(search).get("filter");
+  useEffect(() => {
+    if (teamNameFilter) {
```
jnu: a couple thoughts on this:
1) the `handleTableSearchFilteredData` function doesn't strictly filter by team name. we might end up with false positives where a team name is a subset of an unrelated dataset name. a low-key way to fix this is add support for structured queries. this doesn't have to be super fancy, a low-key way to do this might look like:
```ts
// Support structured queries like "team:foo"
const queryParts = searchText.toLowerCase().split(/:(.*)/, 2);
// All the columns we allow searching on
const allColumns = ["dataset", "team", "program"];
// If it's not a structured query, add a placeholder empty string
if (queryParts.length < 2) {
  queryParts.unshift("");
}
// Make sure that the parsed column is a real column. If it's not,
// revert the parsing to treat it as though no column was given.
if (queryParts[0] && !allColumns.includes(queryParts[0])) {
  queryParts[1] = queryParts.join(":");
  queryParts[0] = "";
}
const [column, query] = queryParts;
// Filter all the source rows
const filtered = data.filter(row => {
  // For structured search, only return matches on the given column
  if (column) {
    return row[column].toLowerCase().includes(query);
  }
  // Otherwise search all columns
  return allColumns.some(col => row[col].toLowerCase().includes(query);
});
```

then we could support things like `?filter=team:foo` or `?filter=program:some%20program`

2) do we ever want the value that a user types in the autocomplete to sync back to the URL `?filter=`? might be useful for sharing queries
3) no need to bail early here if the team name filter is missing, especially if you bail early in the `handleTableSearchFilteredData` function on the empty string.

client/src/pages/Home/Home.tsx
```diff
@@ -135,14 +72,14 @@ const Home = (): JSX.Element => {
...
-  const rowData = getTableData(originalTeamData.slice(), t);
+  const allTableData = getTableData(originalTeamData.slice(), t);
+  const [filteredData, setFilteredData] =
+    useState<Array<TableData>>(allTableData);
 
   // Filters datasets table by search term
-  const handleTableSearchFilter = (searchText: string) => {
-    const data = [...rowData];
+  const handleTableSearchFilteredData = (searchText: string) => {
+    const data = [...allTableData];
```
jnu, line 109: looks like `searchText` ought to be `toLowerCase`d here

jnu: also you can also return early if `searchText` is the empty string since `includes` will always be true

client/src/layout/AppSidebar.tsx
```diff
@@ -63,16 +68,28 @@ export const AppNormalUserSidebarMenu = () => {
...
+            }) => {
               return (
-                <Menu.ItemGroup key={program.key} title={program.team}>
+                <Menu.ItemGroup
+                  key={program.key}
+                  title={
+                    <Link
+                      to={{
+                        pathname: `/`, // go to user homepage
+                        search: `?filter=${program.team}`,
```
jnu: you should call `encodeURIComponent` on the team name. in particular if the team name contains an `&` this will not behave as expected.

(see also my structured search comment elsewhere. another option would be to have different query params, like `?team=${encodeURIComponent(program.team)}`)

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+
+  return (
+    <Table<TableData>
+      loading={loading}
+      dataSource={filteredData.length > 0 ? filteredData : rowData}
+      columns={columns}
+      rowKey={(dataset) => dataset.id}
+      footer={() =>
+        filteredData.length > 0
+          ? `Showing ${filteredData.length} of ${rowData.length} results`
```
jnu: can these come from i18n translations?

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+              <Button type="primary" icon={<PlusOutlined />}>
+                {t("addData")}
+              </Button>
+            </Link>
+            <Link
+              to={{
+                pathname: `/dataset/${datasetId}/details`,
+              }}
+            >
+              <Button icon={<InfoCircleOutlined />}>View Details</Button>
```
jnu: can this use i18n?

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+              {tag.toUpperCase()}
+            </Tag>
+          );
+        });
+      },
+    },
+    {
+      dataIndex: "id",
+      width: 250,
+      render: function btn(datasetId: string) {
```
jnu: just curious, why is this not an arrow function? (if it's due to the name linting, i usually move it out to a separate `const` which i think makes the code more readable anyway)

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+    {
+      title: "Tags",
+      key: "tags",
+      dataIndex: "tags",
+      width: 250,
+      render: (tags: string[]) => {
+        return tags.map((tag: string) => {
+          const color = "blue";
+          return (
+            // TODO: Create component to link tags to datasets with the same tags
```
jnu: is this tracked somewhere?

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+            <Tag color={color} key={tag}>
+              {tag.toUpperCase()}
```
jnu: curious why we're upper-casing tags? just fyi tags are normalized on the backend to all have capitalized first letters

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,123 @@
...
+          </Space>
+        );
+      },
+    },
+  ];
+
+  return (
+    <Table<TableData>
+      loading={loading}
+      dataSource={filteredData.length > 0 ? filteredData : rowData}
```
jnu: hm - so if there are no matches for a query we show everything? that seems potentially odd to me

client/src/pages/Home/HomeSearchAutoComplete.tsx
```diff
@@ -11,18 +10,23 @@ interface AutoCompleteProps {
 const HomeSearchAutoComplete = ({
   onSearch,
 }: AutoCompleteProps): JSX.Element => {
-  const { t } = useTranslation();
+  const searchBarLabel = "Search team and dataset";
```
jnu: can this come from i18n?

client/src/pages/Home/HomeSearchAutoComplete.tsx
```diff
@@ -11,18 +10,23 @@ interface AutoCompleteProps {
...
-      style={{ width: 300 }}
-      aria-label="Search"
-      placeholder={`Search your ${t("program")}s`}
-      aria-expanded="false"
-      onSearch={onSearch}
-    >
-      <Search allowClear />
-    </AutoComplete>
+    <>
+      <label htmlFor="Search">
```
jnu, line 19: for accessibility, the label's `for` attribute needs to have the value of an `id` of an element (maybe there's a `#Search` element here but and im missing it?)

jnu (changes requested):
made a few suggestions which will hopefully clean up searching a bit. overall looks great, nice work!!

client/src/pages/Home/Home.tsx
seerocode: Good points here! What if I do a followup PR for this instead right after this is merged?

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: They were uppercase before the backend was changed to capitalize all letters. Also looks better UI wise IMO.

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: Yeah this is something I'm fixing (added as a TODO here). The dataset query doesn't load until after the component is rendered.

client/src/pages/Home/HomeSearchAutoComplete.tsx
seerocode: Didn't miss it but I did. I'll add the id to the auto complete wrapper.

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: Yeah this was already being tracked here: https://app.clubhouse.io/stanford-computational-policy-lab/story/163/decide-how-to-present-tags

Deleted comment since the story exists.

client/src/App.css
```diff
@@ -15,3 +15,13 @@
   padding-left: 0;
   padding-right: 0;
 }
+
+.visually-hidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
```
nmanduj1, line 21: fancy css- ive never seen "clip" before!

nmanduj1 (approved):
looks good to me! 🚢

client/src/pages/Home/Home.tsx
seerocode: @jnu I added this comment to the structured search ticket so it can be worked on with this in mind: https://app.clubhouse.io/stanford-computational-policy-lab/story/301/structured-search-in-filter-bar

In the meantime, since the param is used to filter the data in the table, I added the data filter as a side effect instead.

client/src/pages/Home/Home.tsx
```diff
@@ -135,53 +74,107 @@ const Home = (): JSX.Element => {
...
+  const [filteredData, setFilteredData] =
+    useState<Array<TableData>>(allTableData);
 
-  const originalTeamData = allTeams?.data?.teams || data?.user?.teams || [];
-  const rowData = getTableData(originalTeamData.slice(), t);
+  // Searches table for team name if selected from user sidebar
+  const { search } = useLocation();
+  const teamNameURLParam = new URLSearchParams(search).get("team");
+  useEffect(() => {
+    if (data && !search) {
```
jnu: you could clean up the code a bit by using explicit return statements, instead of re-evaluating conditions like `data`. for example:

```ts
if (!data) {
  return;
}

if (!search) {
  setFilteredData(allTableData);
  return;
}

if (teamNameURLParam) {
  const filter = ...;
  setFilteredData(filter);
  return;
}
```

that way the code doesn't have to keep evaluating redundant and/or unnecessary conditions. it also makes the code a little easier to read for humans, since you don't have to keep all the different states in mind while reading to the end of the function. i would guess this helps avoid errors in refactoring later.

client/src/pages/Home/Home.tsx
```diff
@@ -135,53 +74,107 @@ const Home = (): JSX.Element => {
...
-  const handleTableSearchFilter = (searchText: string) => {
-    const data = [...rowData];
+  const handleTableSearchFilteredData = (searchText: string) => {
+    const data = [...allTableData];
     const filteredData = data.filter(({ team, dataset }) => {
       team = team.toLowerCase();
       dataset = dataset.toLowerCase();
-      return team.includes(searchText) || dataset.includes(searchText);
+      return (
+        team.includes(searchText.toLowerCase()) ||
```
jnu: slightly better to store `searchText.toLowerCase()` in a variable so you don't have to do the operation multiple times

client/src/pages/Home/Home.tsx
```diff
@@ -135,53 +74,107 @@ const Home = (): JSX.Element => {
...
 
-  if (error) return <ErrorFallback error={error} />;
+  const renderFilterAlert = (): JSX.Element => {
+    const alertPrefix = `${t("user.homePage.showingDatasetsFor", {
+      term: "team",
+    })}: `;
+    const alertSuffix = <Text strong>{teamNameURLParam}</Text>;
+
+    const message = !teamNameURLParam ? (
+      <>{t("missingTeamParam")}</>
```
jnu: is there a reason this text has to be wrapped in a fragment? (usually you can use text directly as a react node, and `<>text</>` is rendered identically to `text`, but maybe something weird is going on with the typings in antd?)

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,126 @@
...
+}: TableProps): JSX.Element => {
+  const { t } = useTranslation();
+  const [searchText, setSearchText] = useState<string>("");
+
+  const highlightTeamName = (text: string) => {
+    teamNameFilterText && setSearchText(teamNameFilterText);
+    return (
+      <Highlighter
+        highlightStyle={{ backgroundColor: "#ffc069", padding: 0 }}
+        searchWords={[searchText.toString()]}
```
jnu: why do we need to call `toString` here? it looks like the type of `searchText` is already a string?

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,126 @@
...
+        autoEscape
+        textToHighlight={text ? text.toString() : ""}
```
jnu: again why is toString necessary? 🤔

client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -1,12 +1,126 @@
...
+        >
+          <Button icon={<InfoCircleOutlined />}>{t("viewDetails")}</Button>
+        </Link>
+      </Space>
+    );
+  };
+
+  const renderTags = (tags: string[]) => {
+    return tags.map((tag: string) => {
+      const color = "blue";
```
jnu: it's better not to hardcode colors like this, for two reasons. first is general DRYness, it's almost always better to define constants rather than have strings scattered around the application, so that it's easier to change later on.

second is that especially for colors, our goal is to make the app themeable. the technique to do that usually consists of two parts: 1) a color palette consisting of specific RGB colors, and 2) a theme that assign semantic value using those colors.

for example:

```ts
const palette = {
  blue: "#2323e8",
  orange: "#e88623",
};

const theme = {
  primary: palette.blue,
  accent: palette.orange,
};
```
then in your UI you reference the theme values, and *not* the colors from the palette directly. for example:

```ts
<Tag color={theme.primary}>{tag}</Tag>
```
this lets us define a custom color palette and/or theme without changing any of the UI code. so we could for example swap out the theme at buildtime, or even at runtime, to make the app fit in with any organization's in-house style.

you don't have to develop the theming capabilities fully now (that would include things like fonts, logos, etc), but we should at least try to avoid having colors defined in random places in the UI, so that it's easier to update later.

jnu (approved):
looks good, made a few stylistic comments but nothing major! nice work!

client/src/pages/Home/Home.tsx
seerocode: thank you! this helps a lot. i'll make the changes.

client/src/pages/Home/Home.tsx
seerocode: Thanks for catching this! I had refactored this earlier and yes this does not need to be wrapped in a fragment. The following false condition does though. I felt splitting up as prefix and suffix made the translation more readable.

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: This is straggler stuff when I was refactoring an ant design highlight example. I'll clean up.

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: Fixed.

client/src/pages/Home/HomeDatasetsListTable.tsx
seerocode: Fair point! I remember this was taken directly from ant design code and it seems they have since updated how they render color tags. I moved the color to a config file for now along with the orange color for text highlight.

PR #110, opened 2021-09-16
Handle shrinking table views

nmanduj1:
https://user-images.githubusercontent.com/6188110/133659616-6010a958-bcba-4fd1-8d81-2c2c3bdf9dca.mov

### DIFF
--- a/client/src/__tests__/Home.test.tsx
+++ b/client/src/__tests__/Home.test.tsx
@@ -1,5 +1,5 @@
 import { ApolloProvider } from "@apollo/client";
-import { act, render, screen, within } from "@testing-library/react";
+import { act, render } from "@testing-library/react";
 import { createMemoryHistory } from "history";
 import { axe } from "jest-axe";
 import { Router } from "react-router-dom";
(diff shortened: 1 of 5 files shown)


client/src/pages/Home/HomeDatasetsListTable.tsx
```diff
@@ -70,37 +82,65 @@ const HomeDatasetsListTable = ({
   };
 
   const columns: ColumnsType<TableData> = [
+    {
+      title: `Team & Dataset`,
+      key: "teamDataset",
+      render: (record) => (
+        <Fragment>
```
seerocode: I think you can also use the short annotation `<> stuff </>` here!

seerocode (approved):
image

nmanduj1:
@seerocode, I ended up removing the logic I had for the responsive tests (tragic but felt was so cumbersome and brittle ! ) and also removed the static tests - it's not optimal to leave this untested but I'm not sure adding responsive tests is our highest priority rn. I know you love tests, so please share your thoughts if you have ideas on how to better handle this!  If you're ok with this removal, I will merge and move on to a more pressing MVP ticket. thank u so much !

client/src/__tests__/Home.test.tsx
```diff
@@ -23,136 +23,140 @@ async function wait(ms = 0) {
...
+//         <Router history={history}>
+//           <Home />
+//         </Router>
+//       </ApolloProvider>
+//     </AuthProvider>
+//   );
+
+//   await wait();
+
+//   screen.getAllByText(/Search team and dataset/i);
```
seerocode: At what point did the tests break? If it's for text search then we can omit static text assertions like `Search team and dataset` since I agree it's brittle to check on that and probably better off with snapshots for that. We can check for the ID instead or add a `test-id` attribute since that is less likely to change.

seerocode: If it helps, I can start a new branch to remove assertions on text and see if that makes a difference in your branch.

seerocode: OK, I think I found the issue. The responsive key in the columns are causing them to be hidden in the DOM when using the react testing library. Seems like a weird thing with ant design and react testing lib.

seerocode: Dug deeper and I think I found the culprit: RTL uses jest with jsdom under the hood which doesn't seem to support pseudo elements. The responsive columns in ant-design use the `::before` pseudo-element which is probably why it's not working with these tests. 

image



Since only the responsiveness columns are an issue, one thing we can do merge this in is to skip the tests with `test.skip(...` and add a comment to explain what's going on.

Here's the issue reported in jsdom: https://github.com/jsdom/jsdom/issues/1928

seerocode: Alternatively, we can change these to snapshot tests and that should cover us for the most part.

PR #116, opened 2021-11-01
Fix error on user update from admin panel

seerocode:
The error in [this comment](https://github.com/stanford-policylab/bbc-50-50/pull/116#discussion_r740392262) is caused by a bug in the current version of fastapi-users we are using. It is [fixed](https://github.com/fastapi-users/fastapi-users/issues/733) in version 8.0.0 but that version uses a newer version of the starlette package which is incompatible with the latest release of ariadne.

Ariadne's next milestone [release](https://github.com/mirumee/ariadne/milestone/13) should have starlette updated and in the meantime, this is a hack to work around the inability to update fastapi-users.

I also tried something more complicated in the API in the last couple of days by creating a new patch users route in the users router in an attempt to override the fastapi-users one and modifying the request in middleware (which is frowned upon). 

This small hack much quicker to undo and doesn't interfere with other functions.

### DIFF
--- a/client/src/services/account.ts
+++ b/client/src/services/account.ts
@@ -96,7 +96,8 @@ const checkError = async (
     case httpStatus.FORBIDDEN:
       throw new Error("FORBIDDEN");
     default:
-      throw new Error("UNKNOWN_ERROR");
+      json = await response.json();
+      throw new Error(json.detail || "UNKNOWN_ERROR");
   }
 };
 
--- a/client/src/__tests__/AdminEditUser.test.tsx
+++ b/client/src/__tests__/AdminEditUser.test.tsx
@@ -183,7 +183,8 @@ it("allows admin to modify user", async () => {
     {
       first_name: "Cat",
       last_name: "Berry",
-      email: "tester@notrealemail.info",
+      // TODO: Uncomment when Ariadne version is updated. See: https://github.com/stanford-policylab/bbc-50-50/pull/116
+      // email: "tester@notrealemail.info",
       roles: ["be5f8cac-ac65-4f75-8052-8d1b5d40dffe"],
       teams: ["472d17da-ff8b-4743-823f-3f01ea21a349"],
     }
--- a/client/src/pages/Admin/EditUser.tsx
+++ b/client/src/pages/Admin/EditUser.tsx
@@ -305,6 +305,20 @@ export const EditUser = () => {
         wrapperCol={{ span: 14 }}
         initialValues={initialFormState}
         onFinish={async (values) => {
+          // TODO: Remove this hack when Ariadne updates starlette dependency for fastapi-users v8 to work
+          if (initialFormState.email === values.email) {
+            const temp_values = { ...values };
+            // Remove email field if it has not changed so the PATCH endpoint doesn't throw a dupe email error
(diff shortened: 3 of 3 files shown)


client/src/services/account.ts
```diff
@@ -96,7 +96,8 @@ const checkError = async (
     case httpStatus.FORBIDDEN:
       throw new Error("FORBIDDEN");
     default:
-      throw new Error("UNKNOWN_ERROR");
+      json = await response.json();
+      throw new Error(json.detail || "UNKNOWN_ERROR");
```
jnu, line 100: just curious, what http status codes are you encountering that this is useful? i'd only expect the `json.detail` to be present on 4xx codes, and the main ones are covered with 400, 401, 403, and 422 already. (maybe it'd be worth adding 404, too?)

seerocode: Not the codes, the response is what I was missing. `UNKNOWN_ERROR` alone was hiding this error 400 message: 
image
 
I believe I was encountering this issue when I was incrementally testing fastapi-users verison bumps. I don't remember the error code because it seems to be fixed now (probably in some intermediate version) but it kept saying unknown error instead of the actual error. 

I felt it was good to see the actual unknown error if some level of detail exists.

jnu: yep i agree it's good to see the actual error! my point is that any error that we expect to happen should be handled by one of the other cases in the switch, not the `default` case. (that's a sign that we haven't fully thought through the error flows, and possibly aren't handling them correctly. the api might even be throwing a `500 - internal server error` here, which could indicate other problems, such as leaking connections or improperly closed transactions.)

in this case, `user email already exists` sounds to me like a `400 - bad request` or a `422 - not processable`, which should be handled by other cases.

jnu: also it may be worth just adding a `console.error` logging line for development to print the actual error to the console (which for various reasons you might not want to do in production or in the UI directly).

seerocode: it's a bad request error which is working now without this change!

nmanduj1 (approved):
Looks good to me! Are the hacks your own or were they suggested as a suitable temporary solution by the interwebs? 

Only suggestion I have: can we add a quick note to the ReadMe specifying that these important todo's exist on the project and must be updated at some point? Might be helpful to some poor soul in the future.

Other than that thought, 🚢🚢🚢🚢🚢  !

seerocode:
@nmanduj1 Added as a story! https://app.shortcut.com/stanford-computational-policy-lab/story/369/remove-hack-for-fast-api-users-patch-endpoint

This is only a problem for folks using ariadne, fastapi-users, and fastapi due to the incompatible starlette package versions across each release. Because it's a very specific combination of packages that are causing the issue, there doesn't seem to be a suitable workaround anyone has mentioned. 

Here's when the first issue was raised and the maintainer noted they are working on it: https://github.com/mirumee/ariadne/issues/693. My guess is everyone else is waiting on that release as well?

This might be why fastapi now recommends `strawberry` as the first option for using graphql with their library because it has docs specific to the fastapi framework: https://fastapi.tiangolo.com/advanced/graphql/

nmanduj1:
Nice ! oo, gotcha! Thanks for explaining- you rock !

PR #118, opened 2021-11-12
[SC359] Add reporting filter and target for category values

seerocode:
Added filters for the query date range and optional target selector (won't be used in MVP).

### DIFF
--- a/api/reporting_queries.py
+++ b/api/reporting_queries.py
@@ -1,8 +1,7 @@
-from operator import and_
 from typing import Any
 from ariadne import convert_kwargs_to_snake_case, ObjectType
 from ariadne.types import GraphQLResolveInfo
-from sqlalchemy.orm.util import outerjoin
+from sqlalchemy.sql.expression import label
 from sqlalchemy.sql.functions import func
 from database import CategoryValue, Entry, Record, Category, Dataset, Tag, Team, Target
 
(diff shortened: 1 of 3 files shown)


seerocode:
My thought here is we can add some dropdown filter to each category section in the overview so admins can select "Cisgender Women" and type in "50"% and the data will change to stats on datasets that meet that global target.

api/reporting_queries.py
```diff
@@ -89,3 +96,52 @@ def resolve_categories_overview(obj: Any, info: GraphQLResolveInfo, **kwargs):
...
+@category_overview.field("categoryDatasets")
+@convert_kwargs_to_snake_case
+def resolve_category_overview_datasets(category: Any, info: GraphQLResolveInfo):
+    '''GraphQL query to fetch overview for all categories
+    :returns: category object
+    '''
+    session = info.context['dbsession']
+
+    filters = list()
+
```
seerocode, line 110: This query is a bit redundant with the one above. This might work better as a fragment but will leave as is for now as it's working!

We can use the result from this query to group by dataset on the frontend and return the datasets that meet the minimum target, etc.

seerocode:
@nmanduj1 this is ready!

seerocode:
Hey @nmanduj1 , is this ok to merge?

nmanduj1 (approved):

